PR #70, opened 2015-08-13
improved error handling in update notifcations transition.

mandric:
Issue: https://github.com/medic/medic-webapp/issues/1144

Refactored `addErr` and `addMsg` so that when passed in bad
configuration we setup an error on the doc and wrote some tests.


garethbowen:
Looks good. Go for merge.

PR #71, opened 2015-08-18
modified validation lib to better support unicode query values.

mandric:
Issue: https://github.com/medic/medic-webapp/issues/1147

Stopped URL encoding parameters before passing to couchdb lucene/fti lib
because otherwise they are encoded twice. Only escaping quotes to
prevent injections.

This validation lib will probably get refactored soon to use
https://github.com/medic/lucene-query-generator but for now wanted to
fix a bug that prevents using `exists('द', 'patient_id')` validation
on a patient visit form.

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -32,10 +32,13 @@ var _exists = function(doc, query, callback) {
 };
 
 var _formatParam = function(name, value) {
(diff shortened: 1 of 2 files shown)


mandric:
Thinking to open an issue for the lucene-query-generator refactor, it won't take long but I didn't want to introduce more complexity for this bug fix by doing a refactor.

PR #73, opened 2015-10-07
added support for outgoing_deny_list conf in default_response transit…

mandric:
outgoing_deny_list is a new configuration that is a comma separated list of strings. If a string in that list matches the beginning of the phone we are setting up a response with a denied state.  The pending message process will ignore these messages and those reports will be left without an auto-reply.   The denied messages still show up in the messages export.

Issue: https://github.com/medic/medic-webapp/issues/750

### DIFF
--- a/transitions/default_responses.js
+++ b/transitions/default_responses.js
@@ -38,8 +38,8 @@ module.exports = {
             return false;
         }
         return _.every(conf.split(','), function(s) {
-            // skip empty strings
-            if (!s) {
+            // ignore falsey inputs
+            if (!s || !doc.from) {
                 return true;
             }
             return !doc.from.match(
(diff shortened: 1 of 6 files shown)


transitions/default_responses.js
```diff
@@ -23,13 +22,32 @@ module.exports = {
...
+     * Return false when the recipient phone matches the denied list.
+     */
+    _isResponseAllowed: function(doc) {
+        var self = module.exports,
+            conf = self._getConfig('outgoing_deny_list') || '',
+            ret = true;
+        if (self._isMessageFromGateway(doc)) {
+            return false;
+        }
+        _.each(conf.split(','), function(s) {
```
garethbowen: It might be nicer to use `_.every` here. It does the short circuit for you.

PR #74, opened 2015-10-27
fixed uniqueWithin validation bugs.

mandric:
Issue: https://github.com/medic/medic-projects/issues/56

This validation was failing because the lucene query was incorrectly
constructed.

- To query on integer values we need to use the <int> prefix on the
  query otherwise the query doesn't match and always returns an empty
  array.

- formatParam was failing on integer values with:

    lib/validation.js:36
    value = value.replace(/"/g, '\\"');
                  ^
    TypeError: Object 1 has no method 'replace'

- It seems the ISO date standard (the one returned by .toISOString()
  with the trailing Z that designates Zulu/GMT time) fails on Lucene.  I
  opened an upstream issue for this and left a comment in the code.

Net change looks like this:

// fails
```
patient_name:"irene" AND last_menstrual_period:"1" AND reported_date:[2015-07-27T13:14:40.564Z TO 3000-01-01T00:00:00]
```

// works
```
patient_name:"irene" AND last_menstrual_period<int>:1 AND reported_date:[2015-07-27T13:14:40.564 TO 3000-01-01T00:00:00]
```

### DIFF
--- a/db.js
+++ b/db.js
@@ -42,6 +42,8 @@ module.exports.user = settings.username;
 module.exports.fti = function(index, data, cb) {
     var path = '/_fti/local' + settings.db 
         + '/_design' + settings.db + '/' + index;
+    logger.debug('fti path: ', path);
+    logger.debug('fti query: ', data);
     client.request({
         path: path,
         query: data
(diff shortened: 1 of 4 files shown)


mandric:
Feel free to merge when it's ready browndav.

mandric:
browndav I was hoping to get this in RC today so let me know if there's anything I can do to help.

browndav:
Code review passed; merge away.

PR #76, opened 2016-06-28
Use new translations doc format. Issue: medic/medic-webapp#1706

garethbowen:
Removed default translations as all the default values were equal to the key and the translation implementation falls back to returning the key anyway.

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,7 +1,9 @@
 var _ = require('underscore'),
     follow = require('follow'),
+    db = require('./db'),
+    logger = require('./lib/logger'),
     config = require('./defaults'),
-    logger = require('./lib/logger');
+    translations = {};
 
 function initInfo(callback) {
     var db = require('./db'),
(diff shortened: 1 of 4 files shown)


config.js
```diff
@@ -74,14 +83,17 @@ module.exports = {
...
+    },
     init: function(callback) {
-        var db = require('./db');
         db.medic.get(config.settings_path, function(err, data) {
             if (err) {
                 return callback(err);
             }
             initConfig(data, callback);
         });
+        loadTranslations();
```
mandric: Same thing here, we don't care (only log an error) if translations don't load?

test/unit/default_responses.js
```diff
@@ -207,6 +205,7 @@ exports['add response if unstructured message and setting enabled'] = function(t
     };
     transition.onMatch({ doc: doc }, {}, {}, function(err, changed) {
         test.ok(messageFn.calledOnce);
+        console.log();
```
mandric: oops

PR #77, opened 2016-07-07
disable transitions by default and harden configuration a bit.

mandric:
Support `disable` but ignore `load` setting, backwards compatible with
old style configuration.  

Issue: https://github.com/medic/medic-webapp/issues/2207

### DIFF
--- a/transitions/index.js
+++ b/transitions/index.js
@@ -6,13 +6,84 @@ var _ = require('underscore'),
     config = require('../config'),
     transitions = {};
 
+/*
+ * All transitions are disabled by default, add new ones here to make it
+ * available to the configuration.  For security reasons, we want to avoid
+ * doing a `require` based on a random input string, hence we maintain this
+ * index transitions.
+ */
+var availableTransitions = [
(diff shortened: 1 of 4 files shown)


mandric:
@garethbowen let me know if you think this is fine as-is or if I should remove the `process.env.TEST_ENV` test and refactor the code so we can test the config/transitions loader.

transitions/index.js
```diff
@@ -29,70 +29,35 @@ var availableTransitions = [
   'update_sent_forms'
 ];
 
-/*
- * Supported configuration:
```
estellecomment: it was really nice to get a list of supported formats! Here or somewhere else, it's useful.

mandric: Ok, I basically converted that documentation into tests. But I think you're right both are useful.  I should probably add something to the actual docs/readme...

mandric: Ok I added configuration examples to the readme, should be all set now. https://github.com/medic/medic-sentinel/blob/2207-disable-transitions-by-default/README.md

PR #78, opened 2016-07-07
Client side ddoc. Issue: medic/medic-webapp#2206

garethbowen:


alxndrsn:
Is there a clear reason why sentinel should be using the `medic-client` ddoc?  Is it for all _clients_?  My initial assumption was that it was more for webapp users, but this may be wrong.

garethbowen:
> Is there a clear reason why sentinel should be using the medic-client ddoc?

The "medic" ddoc is _only_ available on the server, and the "medic-client" ddoc is available on both the client and the server. I didn't want to include the view in both "medic" and "medic-client" as this is more bloat and also creates more views that have to be updated on change. So I moved the views that were required on the client to "medic-client" and modified all callers to use the new ddoc.

PR #79, opened 2016-07-13
v0 v2 compatibility

mandric:
Merged `develop` onto `master` here.  Not really sure how stable this branch is, but tests are passing.  I'm thinking we merge this to `master` and keep using `develop` for v2 (or rename the `develop` to `v2`) merging continuously into master until we have more time to test and consider the merge stable.  Also since sentinel on v2 production instances is mostly disabled this should not pose too much risk.  Patching sentinel on v0 should not be an issue because we can still branch and patch from the current submodule commit until it is EOL.

`git log --graph` is a bit ugly depending on your taste.  Will need feedback here from the team if this is acceptable or if there is a better/preferred way to get these two trees merged.


mandric:
duplicate/deprecated

PR #80, opened 2016-07-13
v0 v2 compatible improved

mandric:
Second attempt at this merge, this is master onto develop.  The changeset looks much better and less risky, and git log --graph also seems ok.  We can probably just start using this in alpha/beta and if v0 needs a patch then we branch from an old commit.
 
This merge also allows transitions to run on data records with XML content type.  We need transitions for XML data records too, depending on the use case.  Also now by default all transitions are disabled, which has the same desired effect but still allows us to enable them when need

### DIFF
--- a/transitions/index.js
+++ b/transitions/index.js
@@ -68,21 +68,16 @@ var canRun = function(options) {
         doc = change.doc,
         transition = options.transition;
 
-    var isRevSame = function(doc) {
+    var isRevSame = function() {
         if (doc.transitions && doc.transitions[key]) {
             return parseInt(doc._rev) === parseInt(doc.transitions[key].last_rev);
         }
         return false;
     };
 
-    var isV2XmlDataRecord = function(doc) {
-        return doc.type === 'data_record' && doc.content_type && doc.content_type === 'xml';
-    };
-
     /*
      * Ignore deleted records, confirm transition has filter function and
      * skip transition if filter returns false.
-     * Ignore 2.x XML data_records, they don't use need sentinel post-processing.
      *
      * If transition was already applied then only run again if doc rev is not
      * equal to transition rev.  If revs are the same then transition just ran.
@@ -95,8 +90,7 @@ var canRun = function(options) {
         transition &&
         typeof transition.filter === 'function' &&
         transition.filter(doc) &&
-        !isRevSame(doc) &&
-        !isV2XmlDataRecord(doc)
+        !isRevSame(doc)
     );
 };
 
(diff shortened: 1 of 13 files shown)


mandric:
I deleted the old `v0-v2-compat` branch but if you wanted to look at that for some reason to compare I can push it back up.

transitions/index.js
```diff
@@ -95,8 +90,7 @@ var canRun = function(options) {
         transition &&
         typeof transition.filter === 'function' &&
         transition.filter(doc) &&
-        !isRevSame(doc) &&
-        !isV2XmlDataRecord(doc)
```
garethbowen: Did we decide to disable all transitions instead of having this check?

transitions/registration.js
```diff
@@ -265,6 +265,7 @@ module.exports = {
             if (err) {
                 callback(err);
             } else if (registrations.length) { // id collision, retry
+                logger.warn('Registration ID ' + id + ' is not unique, retrying...');
```
garethbowen: Make this an info or debug.

transitions/index.js
mandric: Yeah develop already has transitions disabled by default, which has the same effect but still allows us to enable them when needed.  We need transitions for XML data records too, depending on the use case.

transitions/registration.js
mandric: I actually used `warn` because this can create issues with sentinel and should be infrequent.  If it starts happening a lot it's highly recommended to increase the id format/length because the transitions are processed serially and this could potentially loop forever.  We actually had a case where sentinel wasn't doing anything else except for stuck in that loop.

garethbowen: Fair enough. It'd be nice to notify admins when the ID space is getting close to full, but that's a problem for another day.

mandric: Yeah there is an issue with more details and probably some ideas for better solutions somewhere, I can dig it up...

mandric: For example there was some work towards monitoring the sentinel log to notify admins when errors are logged...

mandric:
@garethbowen now that I think about it I doubt this branch can be compatible with v0 mainly because of translations, right?

garethbowen:
Yeah that's definitely a problem. You could write some code to fallback if the messages docs aren't found. Also with the ddoc split sentinel will be requesting "medic-client" views which don't exist.

mandric:
Maybe we just keep a `v0.4` branch around or something...

garethbowen:
Not a branch. We probably don't even need a tag. webapp depends on sentinel via a submodule which is effectively a tag anyway. We can always branch later if we need to patch an old instance.

mandric:
The only risk there is the ugly git log graph syndrome. 😄

PR #81, opened 2016-07-26
Replace backlog processor. Issue: medic/medic-webapp#2521

garethbowen:


transitions/index.js
```diff
@@ -242,18 +246,62 @@ var applyTransitions = function(options) {
...
-    var backlog_delay = 500; // ms
-    var progress_interval = 500; // items
+    var processed = 0;
+    var changeQueue = async.queue(function(task, callback) {
+        processChange(task, function() {
+            _.delay(callback, PROCESSING_DELAY);
+        });
+    });
 
     var match_types = [
```
mandric: I know you didn't change anything here, but this is better as a constant at the top.

transitions/index.js
```diff
@@ -262,127 +310,57 @@ var attach = function() {
...
+            if (err) {
+                logger.error('backlog: fetch failed for %s (%s)', id, err);
+                return callback();
+            }
+            if (match_types.indexOf(doc.type) === -1) {
+                // not a doc we know how to transition - ignore
+                return callback();
+            }
+            if (processed > 0 && (processed % PROGRESS_REPORT_INTERVAL) === 0) {
+                logger.info('backlog: %d items processed', processed);
```
mandric: So this can be a little confusing, because at some point the backlog has been processed and we start processing the current changes.  We might even remove the concept of "backlog" because there's only one change processor.  It's starts at 0 and records as it goes.  I think you can really remove the idea of "backlog processor" and just have a change processor that starts at whatever is in `sentinel-meta-data` and states that "starting at update_seq x", and "processed x changes" every so often.  That might simplify things.

PR #85, opened 2016-09-26
Patient centric sms workflow. Issue: medic/medic-webapp#2700

garethbowen:


transitions/registration.js
```diff
@@ -181,59 +181,39 @@ module.exports = {
...
+        add_birth_date: function(options, cb) {
+            module.exports.setBirthDate(options.doc);
             cb();
         },
-        assign_schedule: function(db, doc, cb) {
-            var args = Array.prototype.slice.call(arguments);
-            if (args.length < 4) {
-                cb('Please specify schedule name in settings.');
+        assign_schedule: function(options, cb) {
+            if (!options.params) {
```
estellecomment: having trouble tracking down where that `params` comes from, but if it's possible to rename it to `schedules` I'd be happy
Edit: ah, probably from the app_settings. Probably not so changeable then.

transitions/registration.js
```diff
@@ -272,5 +252,43 @@ module.exports = {
                 callback();
             }
         });
+    },
+    addPatient: function(options, callback) {
+        var doc = options.doc,
+            db = options.db,
+            audit = options.audit,
+            id = doc.patient_id,
+            patientNameField = (options.params && options.params.length && options.params[0]) || 'patient_name';
```
estellecomment: `params` again, it's everywhere that one, what's `params[0]`?

transitions/registration.js
```diff
@@ -272,5 +252,43 @@ module.exports = {
...
+            id: id
+        }, function(err, registrations) {
+            if (err) {
+                return callback(err);
+            }
+            if (registrations.length) {
+                // patient already registered, no action required
+                return callback();
+            }
+            db.medic.view('medic-client', 'people_by_phone', {
```
estellecomment: can you create vars for constants instead of using values directly?

transitions/registration.js
estellecomment: create var for `patient_name` value

transitions/registration.js
```diff
@@ -272,5 +252,43 @@ module.exports = {
...
+                include_docs: true
+            }, function(err, result) {
+                if (err) {
+                    return callback(err);
+                }
+                var contact = _.result(_.first(result.rows), 'doc');
+                // create a new patient with this patient_id
+                var patient = {
+                    name: doc.fields[patientNameField],
+                    parent: contact && contact.parent,
```
estellecomment: creating a `person` without a patient is not cool. We should add least log an error, and possibly not create the patient at all (if there's no parent the reminders are probably going to get lost anyway)

transitions/registration.js
```diff
@@ -272,5 +252,43 @@ module.exports = {
                 callback();
             }
         });
+    },
+    addPatient: function(options, callback) {
+        var doc = options.doc,
+            db = options.db,
+            audit = options.audit,
+            id = doc.patient_id,
```
estellecomment: rename to `patientId`, by the end of the function I've forgotten what this id is

transitions/registration.js
abbyad: > creating a `person` without a patient is not cool

@estellecomment, do you mean "creating a `person` without a _parent_ is not cool"?

estellecomment: Ah yes thanks for the catch @abbyad :)

test/unit/registration.js
```diff
@@ -49,3 +65,94 @@ exports['bool expr is ignored (returns true) if not a string or empty'] = functi
...
+
+exports['add_patient trigger creates a new patient'] = function(test) {
+    var change = { doc: {
+        form: 'R',
+        patient_id: '05649',
+        reported_date: 53,
+        from: '+555123',
+        fields: { patient_name: 'jack' }
+    } };
+    var view = sinon.stub().callsArgWith(3, null, { rows: [ { doc: { parent: { _id: 'papa' } } } ] });
```
garethbowen: Commented.

test/unit/registration.js
```diff
@@ -49,3 +65,94 @@ exports['bool expr is ignored (returns true) if not a string or empty'] = functi
...
+        test.equals(getRegistrations.callCount, 1);
+        test.equals(getRegistrations.args[0][0].id, '05649');
+        test.equals(view.callCount, 1);
+        test.equals(view.args[0][0], 'medic-client');
+        test.equals(view.args[0][1], 'people_by_phone');
+        test.deepEqual(view.args[0][2].key, [ '+555123' ]);
+        test.equals(view.args[0][2].include_docs, true);
+        test.equals(view.args[0][0], 'medic-client');
+        test.equals(saveDoc.callCount, 1);
+        test.equals(saveDoc.args[0][0].name, 'jack');
```
garethbowen: I don't think I am. The one below checks that a custom field name can be given as a parameter. This one is checking the default name ("patient_name")

test/unit/registration.js
```diff
@@ -49,3 +65,94 @@ exports['bool expr is ignored (returns true) if not a string or empty'] = functi
...
+    var getRegistrations = sinon.stub(utils, 'getRegistrations').callsArgWith(1, null, [ { _id: 'xyz' } ]);
+    transition.onMatch(change, db, audit, function() {
+        test.equals(getRegistrations.callCount, 1);
+        test.equals(getRegistrations.args[0][0].id, '05649');
+        test.equals(saveDoc.callCount, 0);
+        test.done();
+    });
+};
+
+exports['add_patient uses given arg as patient name field name'] = function(test) {
```
garethbowen: Rewritten...

transitions/registration.js
```diff
@@ -154,17 +154,17 @@ module.exports = {
...
                     var obj = _.defaults({}, doc, doc.fields);
                     if (self.isBoolExprFalse(obj, event.bool_expr)) {
                         return;
                     }
+                    var options = { db: db, audit: audit, doc: doc };
+                    if (event.params) {
+                        // params setting get sent as array
+                        options.params = event.params.split(',');
+                    }
                     series.push(function(cb) {
```
garethbowen: Agreed. Didn't want to update any more than necessary. `cb` is used everywhere here  :(

transitions/registration.js
```diff
@@ -181,59 +181,39 @@ module.exports = {
...
-                return;
-            }
+        add_patient_id: function(options, cb) {
             // if we already have a patient id then return
-            if (doc.patient_id) {
+            if (options.doc.patient_id) {
                 return;
             }
-            self.setId({db: db, doc: doc}, cb);
+            module.exports.setId(options, cb);
```
garethbowen: Again, I totally agree. I'd rather leave it for now until we get around to cleaning up sentinel code more thoroughly. There are many many things I'd change if I start pulling that thread now.

transitions/registration.js
garethbowen: Done.

transitions/registration.js
garethbowen: As above, event.params is a string. For passing multiple params (used in other triggers) you can comma separate them. On line 164 we see `options.params = event.params.split(',')`, which means options.params is an array, even if it doesn't contain any commas (as in this trigger). Again, needs a refactor one day...

garethbowen: I tried that but it's really hard to name well. Used `_.first` to clean up the code instead.

transitions/registration.js
garethbowen: I'd actually find that much more difficult to use. If I'm reading this code I want to know what view it's querying, not the name of a constant.

transitions/registration.js
garethbowen: Actually no the reminders won't be lost. The reminders, validations, automatic responses, etc are based on the report _not_ the patient/contact/parent. So the app will continue to function perfectly.

I thought for a while about not creating the patient at all, but doing it here still gives admins the chance to see the patient information grouped together and makes the problem visible. It also makes it possible for managers to assign the patient to a CHP.

Furthermore I expect most instances will be configured to reject submissions from unknown sources so those submissions will never get here.

transitions/registration.js
estellecomment: hmmm. The point is to make it easier to spot and modify the strings you depend on, later. 
But! This probably is a java habit that makes less sense in javascript, since in javascript all names of everything are potentially typo-able, so strings don't really deserve special treatment. Interesting. Ok!

transitions/registration.js
estellecomment: Hmmmwokay I guess. (you gave it some thought, it's not an omission, which is the main thing I'm checking!)

Where does the patient show up if it's an orphan? Does the new People tab design still allow to see them? (there are less things displayed in the RHS now) Does it also depend on permissions? 
Not necessarily blocker for here, but if it's not clear please file an separate issue to figure out the UX flow of orphan patients, as part of People tab redesign.

PR #88, opened 2016-11-30
Lucene support for CouchDB 2.0 medic/medic-webapp#2935

SCdF:


SCdF:
This is no longer relevant, and was superseded by other PRs.

PR #91, opened 2016-12-07
conditional_alerts fixes for 2.9

mandric:
Applied the diff between conditional_alerts 0.4.x and 2.9.x branches. I'm not sure exactly why the _hasRun logic was removed in 0.4? I'm guessing there was a bug with that RegExp and certain form codes like द. I could not find anything in the git logs but it might be good to track down the history and apply all the changes to the 2.9 branch.  In particular I should check for tests.

Related Issues:
https://github.com/medic/medic-projects/issues/1133
https://github.com/medic/medic-webapp/issues/2978


mandric:
Closing this since the newer one was applied.

PR #92, opened 2016-12-08
Rerun transitions on change if first run failed. medic/medic-webapp#2978

garethbowen:


garethbowen:
It looks like this was fixed in https://github.com/medic/medic-sentinel/commit/c5c4a9908f7c21d7d3b52c6af00222819432f986 so this PR isn't required.

PR #93, opened 2016-12-08
fix conditional alerts part deux

mandric:
fixed bug, don't short ciruit transitions if one errors.
    
    Do not return errors when applying transitions because that will short
    circuit async.series so transitions will stop being applied.
    
    Transitions are async and should return true if the document has
    changed.  If a transition errors then log the error, but don't return it
    because that will stop the series and the other transitions won't run.
    A doc can be modified with error messages but will only get saved to the
    doc if there are other (non-error) changes too.
    
    Some time ago errors messages were added to the doc because there was no
    way to surface parser errors.  Since the configuration in templates and
    conditionals/expressions involves evaluating code, it's useful to
    provide some clue to the user when this fails.  This isn't perfect but
    it's a middleground solution.

Related Issues:
https://github.com/medic/medic-projects/issues/1133
https://github.com/medic/medic-webapp/issues/2978

### DIFF
--- a/transitions/conditional_alerts.js
+++ b/transitions/conditional_alerts.js
@@ -9,20 +9,11 @@ module.exports = {
     _getConfig: function() {
         return _.extend({}, config.get('alerts'));
     },
-    _hasConfig: function(doc) {
-        var self = module.exports;
-        // confirm the form is defined on a reminder config
-        return _.find(self._getConfig(), function(obj) {
-            return obj.form &&
-                doc.form.match(new RegExp('^\\s*'+obj.form+'\\s*$','i'));
-        });
-    },
     _runCondition: function(condition, context, callback) {
         try {
             callback(null, vm.runInNewContext(condition, context));
         } catch(e) {
-            var lines = e.message.split('\n');
-            callback(lines[lines.length - 1]);
+            callback(e.message);
         }
     },
     _evaluateCondition: function(doc, alert, callback) {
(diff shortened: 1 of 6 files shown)


transitions/conditional_alerts.js
```diff
@@ -48,20 +48,12 @@ module.exports = {
...
-            doc.transitions.conditional_alerts
-        );
-    },
     filter: function(doc) {
         var self = module.exports;
         return Boolean(
             doc &&
             doc.form &&
             doc.type === 'data_record' &&
-            self._hasConfig(doc) &&
```
garethbowen: Nice catch - without this any unconfigured reports will rerun every time they change!

transitions/index.js
```diff
@@ -214,40 +220,33 @@ var applyTransition = function(options, callback) {
...
-            change.id, change.seq, key
+            'finished transition %s for seq %s doc %s is %s',
+            key, change.seq, change.id, changed ? 'changed' : 'unchanged'
         );
-        // todo?
-        // prop = doc.transitions && doc.transitions[key] ? doc.transitions[key] : {};
-        //_setProperty('tries', prop.tries ? prop.tries++ : 1)
-        //_setProperty('couch', 'uuid');
-        //
         if (err || changed) {
```
garethbowen: We shouldn't even store `err` - the `hasRun` check isn't sophisticated enough to distinguish between successes and failures. Change this to just `if (changed) {`

garethbowen: It hurts in this case because if a transition that returns an error is followed by a different transition that doesn't have an error then the first transition will never run again. This is exactly the case this PR is trying to solve. We could change all `hasRun` functions to check that `ok=true`, or not store the error. I'll have a look.

mandric: I still think each transition should be free to deal with the info available on `doc.transitions.transition_name` object as it sees fit since it all depends what the transition does. So it could be useful to know error state. We can discuss more later.

PR #94, opened 2017-01-05
Remove timestamps in logs, gardener does it already. #3002

estellecomment:
medic/medic-webapp#3002

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -1,9 +1,2 @@
-var winston = require('winston');
-
-var logger = new (winston.Logger)({
-    transports: [
-        new (winston.transports.Console)({'timestamp':true})
-    ]
-});
-
+var logger = require('winston');
 module.exports = logger;
--- a/server.js
+++ b/server.js
@@ -4,7 +4,7 @@ var config = require('./config'),
 
 if (arg === 'debug') {
     logger.info('setting loglevel to %s.', arg);
-    logger.transports.console.level = arg;
+    logger.transports.Console.level = arg;
 }
 
 if (process.env.TEST_ENV) {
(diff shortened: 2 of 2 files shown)


lib/logger.js
```diff
@@ -2,7 +2,7 @@ var winston = require('winston');
 
 var logger = new (winston.Logger)({
     transports: [
-        new (winston.transports.Console)({'timestamp':true})
+        new (winston.transports.Console)()
```
garethbowen: I think this is the default configuration so you probably don't need to provide anything. Try replacing this with... `var logger = new (winston.Logger);` or something?

alxndrsn: Those brackets look weird to me.  Would `new winston.transports.Console()` be incorrect?

server.js
```diff
@@ -4,7 +4,7 @@ var config = require('./config'),
 
 if (arg === 'debug') {
     logger.info('setting loglevel to %s.', arg);
-    logger.transports.console.level = arg;
```
alxndrsn: Is this change related to date duplication?

alxndrsn:
Looks fine, but:

* I don't understand how the `console`/`Console` typo is related
* I suspect the log output will still duplicate the level, e.g.

### Previous

```2016-12-08T17:19:35.878Z - info: 2016-12-08T17:19:35.877Z - info: startup complete.```

### Fixed

```2016-12-08T17:19:35.878Z - info: info: startup complete.```

^^ will `info` appear twice?

File: server.js (hunk 2 above)
Comment (estellecomment) [inline server.js]: Weirdly, yes. 
Using the default logger, in `lib/logger`, seems to create the `logger.transports` object differently, and `console` is undefined, whereas `Console` is there. 
If you create the logger with arguments (as was done up to now) you get `console`.

File: server.js [hunk 3]
```diff
@@ -20,8 +20,8 @@ config.init(function(err) {
     }
     logger.info('loaded config.');
     if (!arg) {
-        logger.transports.console.level = config.get('loglevel');
-        logger.debug('loglevel is %s.', logger.transports.console.level);
```
alxndrsn: ?  I didn't just make this comment...

alxndrsn: It was 5 days ago

estellecomment:
@alxndrsn your comments about log levels make sense, I agree it would be better, but I'm not finding the time to get to this, so I'm just going to merge. Let me know if you actually really care. 

Is the Console thing clear?

estellecomment:
yeah, one comes from gardener, the other from api/sentinel. 
Cool, thanks!

PR #95, opened 2017-01-18
Add wrapper around felix's view query func, to deal with buggy case w…

estellecomment:
…here both err and result are undefined. medic/medic-projects#1178

### DIFF
--- a/db.js
+++ b/db.js
@@ -52,7 +52,11 @@ var makeDb = function(client) {
     db.view = function(design, view, query, cb) {
         return nativeViewFunc.call(db, design, view, query, function(err, data) {
             if (!err && !data) {
-                return cb(new Error('Both err and data are undefined'));
+                var errMessage = 'Both err and data are undefined in view call.\n' +
+                    'design : ' + design + '\n' +
+                    'view : ' + view + '\n' +
+                    'query : ' + JSON.stringify(query, null, 2) + '\n';
+                return cb(new Error(errMessage));
             }
             return cb(err, data);
         });
@@ -61,7 +65,11 @@ var makeDb = function(client) {
     db.getDoc = function(id, doc, cb) {
         return nativeGetDocFunc.call(db, id, doc, function(err, doc) {
             if (!err && !doc) {
-                return cb(new Error('Both err and doc are undefined'));
+                var errMessage = 'Both err and data are undefined in getDoc call.\n' +
+                    'db : ' + JSON.stringify(db, null, 2) + '\n' +
+                    'id : ' + id + '\n' +
+                    'doc : ' + JSON.stringify(doc, null, 2) + '\n';
+                return cb(new Error(errMessage));
             }
             return cb(err, doc);
         });
--- a/db.js
+++ b/db.js
@@ -37,10 +37,32 @@ var client = couchdb.createClient(
     settings.username,
     settings.password
 );
-module.exports = client.db(settings.db);
+var db = client.db(settings.db);
+
+// Fix for 0.4 : https://github.com/medic/medic-projects/issues/1178#issuecomment-273550990
+var nativeViewFunc = db.view;
+db.view = function(design, view, query, cb) {
+    return nativeViewFunc.call(db, design, view, query, function(err, data) {
+        if (!err && !data) {
+            return cb(new Error('Both err and data are undefined'));
+        }
+        return cb(err, data);
+    });
+};
+var nativeGetDocFunc = db.getDoc;
+db.getDoc = function(id, doc, cb) {
+    return nativeGetDocFunc.call(db, id, doc, function(err, doc) {
+        if (!err && !doc) {
+            return cb(new Error('Both err and doc are undefined'));
+        }
+        return cb(err, doc);
+    });
+};
+
+module.exports = db;
 module.exports.user = settings.username;
 module.exports.fti = function(index, data, cb) {
-    var path = '/_fti/local' + settings.db 
+    var path = '/_fti/local' + settings.db
         + '/_design' + settings.db + '/' + index;
     logger.debug('fti path: ', path);
     logger.debug('fti query: ', data);
--- a/db.js
+++ b/db.js
@@ -24,20 +24,24 @@ if (process.env.COUCH_URL) {
 	}
 } else if (!process.env.TEST_ENV) {
     logger.error(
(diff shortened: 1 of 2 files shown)


db.js
```diff
@@ -37,10 +37,32 @@ var client = couchdb.createClient(
     settings.username,
     settings.password
 );
-module.exports = client.db(settings.db);
+var db = client.db(settings.db);
+
+// Fix for 0.4 : https://github.com/medic/medic-projects/issues/1178#issuecomment-273550990
```
mandric: How about some english here instead of a link? I think if you're going to comment, then comment otherwise I don't want to wade through complicated issues.  Give me your succinct wisdom please!

db.js
```diff
@@ -37,10 +37,32 @@ var client = couchdb.createClient(
...
+var nativeViewFunc = db.view;
+db.view = function(design, view, query, cb) {
+    return nativeViewFunc.call(db, design, view, query, function(err, data) {
+        if (!err && !data) {
+            return cb(new Error('Both err and data are undefined'));
```
mandric: Do the calling functions still throw when they get this?  Just want to make sure we're still not crashing when this is called.

db.js
```diff
@@ -37,10 +37,32 @@ var client = couchdb.createClient(
     settings.username,
     settings.password
 );
-module.exports = client.db(settings.db);
+var db = client.db(settings.db);
+
+// Fix for 0.4 : https://github.com/medic/medic-projects/issues/1178#issuecomment-273550990
+var nativeViewFunc = db.view;
+db.view = function(design, view, query, cb) {
+    return nativeViewFunc.call(db, design, view, query, function(err, data) {
```
mandric: Also don't think you need `.call` here unless you're trying to set a specific context?  You can just do `nativeViewFun(...)`.

mandric: Doh, I see how you're using that now.

estellecomment: ahaaaa yes I doooooooooo 😎  😎  😎

db.js
mandric: Also would be good to return the query details or some data that might help us reconstruct the failed queries.

mandric:
All the crashes point over here https://github.com/felixge/node-couchdb/blob/master/lib/couchdb.js#L185.  But that still doesn't solve the main problem of how to handle reprocessing the change that failed.  In the end backlog processing is needed for sentinel to work.  So patching sentinel any other way doesn't really solve the problem.

db.js
estellecomment: added in comment above.

estellecomment:
Yeah, reprocessing the change that failed will be solved by upgrading to 2.x.

mandric:
Well, the same "problem" exists in master/2.x sentinel.  So the way we handle our db library calls might be worth looking into.  I don't think crashing is really the problem, because the process gets restarted and the backlog processing kicks in.  But crashing specifically this way on an undefined error is not ideal.  Even if we decide crashing is the right thing to do because it's the best way to handle the case where the db library does nothing for us (consider file bug upstream), we should still throw something more descriptive/explicit, and maybe log the info about the failed request? 

From beta.dev sentinel log:
```
2017-01-20T17:30:59.404Z - info: 2017-01-20T17:30:59.403Z - info: saved changes on doc 30e68a9de589d827e5a802d9e7f555b7 seq 897855
 
2017-01-20T17:30:59.999Z - error: 2017-01-20T17:30:59.998Z - error: Error updating metaData Error: socket hang up
    at createHangUpError (_http_client.js:215:15)
    at Socket.socketOnEnd (_http_client.js:293:23)
    at Socket.emit (events.js:129:20)
    at _stream_readable.js:908:16
    at process._tickCallback (node.js:355:11)
 
2017-01-20T17:31:01.514Z - error: /srv/storage/gardener/data/working_dir/aHR0cDovL2xvY2FsaG9zdDo1OTg0L21lZGljL19kZXNpZ24vbWVkaWMvbWVkaWMtc2VudGluZWw=/node_modules/medic-sentinel/transitions/update_clinics.js:87
                if (!data.rows.length) {
                         ^
TypeError: Cannot read property 'rows' of undefined
    at /srv/storage/gardener/data/working_dir/aHR0cDovL2xvY2FsaG9zdDo1OTg0L21lZGljL19kZXNpZ24vbWVkaWMvbWVkaWMtc2VudGluZWw=/node_modules/medic-sentinel/transitions/update_clinics.js:87:26
    at Request._callback (/srv/storage/gardener/data/working_dir/aHR0cDovL2xvY2FsaG9zdDo1OTg0L21lZGljL19kZXNpZ24vbWVkaWMvbWVkaWMtc2VudGluZWw=/node_modules/medic-sentinel/node_modules/nano/lib/nano.js:199:16)
    at self.callback (/srv/storage/
```

estellecomment:
Added more logging. 

Would you rather crash than pass the error? I don't have an opinion on crashing on purpose vs. passing error to callback. I can change to crash if you like.

mandric:
Well, I think passing the error will crash when it gets passed back.  I think handling it like an error is probably the right thing.  Would you agree?  Also do you have an example of what the log output (and potential stack trace) looks like by chance?

estellecomment:
Example of stack trace, when trying to get the ddoc but no couch server is found:
```
error: Failed to create design document: Error: Error: connect ECONNREFUSED 127.0.0.1:5984
error: Failed to create design document: Error: Both err and data are undefined in getDoc call.
db : {
  "name": "/medic",
  "client": {
    "host": "localhost",
    "port": "5984"
  },
  "user": "admin"
}
id : _design/kujua-sentinel
rev : undefined
attachments : undefined
```

estellecomment:
Agreed, handling the error through the callback chain sounds good. 
Ready for rereview!

mandric:
I'm hesitant to put this change into production on 0.4 because it wraps all the sentinel db calls and all it accomplishes is better error messages, and we don't want to spend time improving the 0.4 branch, should probably limit to bug fixes only, and this isn't really a fix...

How do you feel about modifying the callbacks in all calls to those db lib functions and just check for `!err && !data` and then 1) log a message and 2) return early so the other transitions can continue processing? I think that is the lightest touch/less prone to cause any unforeseen issues...

mandric:
Closing this but leaving the branch around for a bit.  @estellecomment you can resubmit against master if you want but we are not accepting these type of improvements to 0.4.x branch because it's bug fixes only.

PR #99, opened 2017-02-08
Patient contact docs use correct id medic/medic-webapp#3084

SCdF:
Not strictly against that ticket, but it was discussed as part of it.
If we do this then it makes them much easier to find later.

### DIFF
--- a/transitions/registration.js
+++ b/transitions/registration.js
@@ -137,9 +137,8 @@ module.exports = {
      * Given a form code and config array, return config for that form.
      * */
     getRegistrationConfig: function(config, form_code) {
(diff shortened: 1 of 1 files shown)


transitions/registration.js
```diff
@@ -137,9 +137,8 @@ module.exports = {
      * Given a form code and config array, return config for that form.
      * */
     getRegistrationConfig: function(config, form_code) {
-        var regex = new RegExp('^\W*' + form_code + '\\W*$','i');
```
SCdF: we didn't need to define this twice, there is a util, which also documents why (ish)

PR #100, opened 2017-02-09
Configurable max sms per message

garethbowen:
medic/medic-webapp#3095

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 node_modules
+logs
+working_dir
--- a/README.md
+++ b/README.md
@@ -19,9 +19,10 @@ Debug mode:
 
 ## Overview
 
-Sentinel uses the changes feed on a CouchDB database and runs a set of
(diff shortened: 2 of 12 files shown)


garethbowen:
@mandric I'm comfortable with it because it'll only happen when truncating messages. With the default truncation limit set to 10 messages this will probably be rare.

mandric:
Also when we make doc structure changes we need to document them somewhere.

garethbowen:
If only we had types... </stefan>

@mandric Seeing as this documentation doesn't exist yet are you alright for me to merge this as is? It's an addition to the doc structure so it shouldn't break any integrations.

mandric:
@garethbowen I was hoping to see at least a sprout of docs before it's merged.  I think that's part of the code review process.

mandric:
I was also wondering a little why doc changes is needed?  I'd like to reduce them as much as possible because it increases footprint everywhere.

PR #101, opened 2017-02-10
Create a patient person if it doesn't exist medic/medic-webapp#3115

SCdF:
Instead of checking if there are any registrations for the patient,
check to see if the patient document exists that we're trying to create,
and if it doesn't create it.

### DIFF
--- a/transitions/registration.js
+++ b/transitions/registration.js
@@ -313,17 +313,15 @@ module.exports = {
             patientId = doc.patient_id,
             patientNameField = _.first(options.params) || 'patient_name';
 
-        utils.getRegistrations({
-            db: db,
-            id: patientId
-        }, function(err, registrations) {
-            if (err) {
+        db.medic.get('patient-' + patientId, function(err, patientContact) {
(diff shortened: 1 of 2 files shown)


transitions/registration.js
```diff
@@ -313,17 +313,15 @@ module.exports = {
             patientId = doc.patient_id,
             patientNameField = _.first(options.params) || 'patient_name';
 
-        utils.getRegistrations({
-            db: db,
-            id: patientId
-        }, function(err, registrations) {
-            if (err) {
+        db.medic.get('patient-' + patientId, function(err, patientContact) {
```
garethbowen: This won't work with any 2.9.x data or anyone migrated from 0.4.x. To rely on this ID format we'd need to write a migration to update patient uuids (which is a pain), but otherwise the ID format is pretty much useless. The alternative is to query the view.

garethbowen: I would pull this out into a utility function which handles the 404 and the ID concatenation.

SCdF:
This PR is no longer relevant, has been superseded by others.

PR #106, opened 2017-02-21
Use translation keys for messages

garethbowen:
medic/medic-webapp#3127


lib/messages.js
```diff
@@ -39,23 +39,29 @@ function addMessage(opts) {
  * Take array of message objects and locale string and return message content.
```
mandric: Probably want to edit this bit of documentation?

lib/messages.js
```diff
@@ -39,23 +39,29 @@ function addMessage(opts) {
...
-        return ret;
+        console.warn('Message property should be an array. Please check that your configuration is valid.');
+        return '';
     }
-    // default to first item in messages array in case locale match fails
-    if (messages.length) {
-        found = _.findWhere(messages, {locale: locale});
-        ret = found ? found.content : messages[0].content;
+    if (!messages.length) {
+        console.warn('Message property array way empty. Please check that your configuration.');
```
mandric: Typos

lib/messages.js
```diff
@@ -39,23 +39,29 @@ function addMessage(opts) {
...
+    if (!configuration) {
+        return '';
+    }
+
+    // use the translation key if provided
+    if (configuration.translation_key) {
+        return utils.translate(configuration.translation_key, locale);
+    }
+
+    // otherwise, use the configured messages
```
mandric: Maybe also stick a note in here about `messages` being deprecated?  Either a comment or a warning?

lib/messages.js
garethbowen: Done.

lib/messages.js
garethbowen: Fixed.

lib/messages.js
garethbowen: Spitting out console warnings is probably a bit aggressive for now. I've updated the comment to indicate it's deprecated. We can add a warning later on when people have had a chance to migrate to the new config.

PR #110, opened 2017-03-02
Removing usage of the _config endpoint medic/medic-webapp#2856

SCdF:
I could not find this export actually being used anywhere, so I just
deleted it.


SCdF:
@alxndrsn this is all that is required to "support" CouchDB 2.0 in Sentinel. Let's just get it merged now to tick it off the list.

alxndrsn:
Weird that the commit references a ticket that's already closed.

PR #111, opened 2017-03-03
Drop doctype whitelist medic/medic-webapp#3166

SCdF:
This allows us to use sentinel for all normal (not deleted or ddoc)
documents.

### DIFF
--- a/transitions/accept_patient_reports.js
+++ b/transitions/accept_patient_reports.js
@@ -13,6 +13,7 @@ module.exports = {
         var self = module.exports;
         return Boolean(
             doc &&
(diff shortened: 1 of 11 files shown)


transitions/accept_patient_reports.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
         var self = module.exports;
         return Boolean(
             doc &&
+            doc.type === 'data_record' &&
```
SCdF: We don't actually need this check because we check for the existence of a form, which is only on a `data_record` anyway. I'm adding it here to be explicit.

PR #114, opened 2017-03-08
Generates patient_id on patients medic/medic-webapp#3000

SCdF:


transitions/generate_patient_id_on_patients.js
```diff
@@ -0,0 +1,11 @@
+var transitionUtils = require('./utils');
+
+module.exports = {
+  filter: function(doc) {
+    return doc.type === 'person' &&
```
mandric: Doesn't this mean all `person` type docs will get a patient_id?  I'm not sure if that's the intent because I don't have enough context but that's pretty odd behavior to me.  Should there be a way to enable/disable which person docs have patient ids?

mandric: We also need to keep in mind that `id` generation is typically limited because they are short by design because they are used in SMS.  So using them where if they are not used will be a waste.  Also other transitions will not complete (sync they all run synchronously in memory on a single change) until this one completes.  So it potentially slows down the transition processing chain.  I'm in favor of having an easy way in configuration to enable/disable which documents are assigned patient ids, similar to how registrations behave by having a boolean expression that is evaluated, and by default it should be off.  There is a cost for generating these types of IDs and we've seen instances in the past that are "stuck" because the entire pool of IDs are in use.

SCdF: > Doesn't this mean all person type docs will get a patient_id? I'm not sure if that's the intent because I don't have enough context but that's pretty odd behavior to me. Should there be a way to enable/disable which person docs have patient ids?

@mandric I tried to answer this in the Github comment linking to this pr:

> NB: this does mean we're going to generate ids for CHWs as well. I'm cool with this (as is Marc). I think honestly it could be useful if people want to be able to submit CHW reports in the future via SMS. The one down-side is that the field is patient_id. Ideally it would be person_id, but it's used in configs, and generally spread throughout the code, so if we're going to fix it I think it's going to be a separate piece of work.

> We also need to keep in mind that id generation is typically limited because they are short by design because they are used in SMS. So using them where if they are not used will be a waste. 

To expand on your concerns, the ratio of patients to CHWs is easily 10:1 or 100:1, so in terms of CHWs adding a strain on the ID system Marc and I didn't think it would be too much of a big deal. It also:
 - Greatly simplifies code and deployment. There is no universal rule where you can take a person in isolation and determine if they are a CHW or not: we would have to have configuration for SMS situations and App situations, and God-knows what when we have both.
 - Can be used in the future to support sending reports about CHWs via SMS. We already support creating reports about CHWs, and we have talked about utilising SMS in a number of ways, so I think this fits.

> Also other transitions will not complete (sync they all run synchronously in memory on a single change) until this one completes. So it potentially slows down the transition processing chain. 

Would this really be a concern? It's literally one write, which occurs if there is a new contact, and that contact was created via enketo.

> I'm in favor of having an easy way in configuration to enable/disable which documents are assigned patient ids, similar to how registrations behave by having a boolean expression that is evaluated, and by default it should be off. 

Right, so this would be the trying to keep it simple thing. Thinking about it though: I guess all we'd need is a couple of extra lines in `onChange` which takes a `bool_expr` configured and runs it against the doc to potentially filter it out? You can't do it in the `filter` function itself unfortunately, but I guess in `onChange` is fine?

mandric: I like the idea of using the `patient_id` as a "short ID" or something that can be used for any person to do various things, that makes perfect sense.  The only problem is generating them is a little expensive, requires querying each time.  Have we discussed potentially changing the IDs to `person-$id` I feel the way we are dealing with `patient_id` currently is pretty brittle.  Adding `patient_id` to every person by default makes it more brittle unfortunately. Thinking...

SCdF: > Have we discussed potentially changing the IDs to person-$id

I have no idea what this means sorry.

> I like the idea of using the patient_id as a "short ID" or something that can be used for any person to do various things, that makes perfect sense.

So I think we should have a call when you're around next. That is already what they're used for. So I presume we're misunderstanding each other.

_Anyway_: the way we generate ids is definitely more complex than simply starting at 1 and heading on up. I don't think it's safe to presume the complexity doesn't come with a purpose. Unfortunately it's not documented anywhere I'm aware of, so we'd need to be very careful here. We'd also need to support both, since you can't change the scheme mid-project.

It's also a completely separate ticket. I don't really want this ticket to be held back because of our interest in maybe making a separate but related part of the system nicer.

SCdF: (having said all that, a reasonably quick fix would be, since ID generation is currently centralised, would be to call the view on startup and then keep the results in memory. That way you don't need use IO to verify collisions. You could even generate N in advance and then just hand them out, if you wanted to split up the time to generate from sentinel walking through the changes feed )

mandric: > > Have we discussed potentially changing the IDs to person-$id

> I have no idea what this means sorry.

I might be conflating, but pretty recently we discussed using a string like `person-$id` where `$id` is something we generate like a short ID, that is used for people's IDs.  So a doc for a person would looks like:

```
{
  "_id": "person-4994821",
  "type": "person",
  ....
}
```

This is a much better approach in my opinion for a few reasons.  1) If we plan to rely on this feature more and more then we should have a valid identifier for people at creation time  2) It's more efficient, and I don't think very difficult to put the ID generation in couchdb/pouchdb

SCdF: @mandric ok yeah we have talked about this. I'm not sure what it has to do with this ticket though?

mandric: After further discussion on hangouts it's determined that it's not necessary/feasible to make changes to how we process the short code or `patient_id` property, so much of the discussion above can be ignored.  But there are still two issues/things to be done to make this PR stable.  The main problem here is that now every person will have a short code or `patient_id` property and we will run out of IDs pretty quickly (maximum at 10000 [still need to confirm](https://github.com/medic/medic-sentinel/blob/master/lib/ids.js)) with a default id format setting of `1111`.  So in order to release this into the wild we'll need to be able to increase that length automatically, that setting will mostly go away (or be moved to the actual transition config only for advanced users), and be managed automatically within sentinel.  Whenever sentinel hits a retry limit (maybe 5 or something) then it will increment the length of that ID automatically by one.  Otherwise sentinel would quickly get stuck on this transition on instances with more than 10k people and all transitions would stop working.

mandric: Also added a functional test for id generation as a minor sanity check because there was no testing around it afaict: https://github.com/medic/medic-sentinel/pull/115  and confirmed independently that a id_format of '1111' does indeed give us a pool of 10000 ids.

SCdF:
@mandric OK I have made the discussed changes. This included changing your test. I can change it back, but it was breaking for me, including with the existing code before I changed it. Give it a read and we can talk about it.

lib/utils.js
```diff
@@ -362,28 +362,29 @@ module.exports = {
         );
     },
     /*
-     * Returns the first document matching the supplied id in the `patient_id`
-     * field.  Optionally matches a form also.
-     */
+    Gets registration documents for the given ids
```
SCdF: I changed this comment because it was wrong: we neither return documents (instead we return rows) nor do we return the first one (we return all of them)

lib/utils.js
```diff
@@ -362,28 +362,29 @@ module.exports = {
...
+        if (ids) {
+            viewOptions.keys = ids;
+        }
+
+        db.medic.view('medic', 'registered_patients', viewOptions, function(err, data) {
             if (err) {
                 return callback(err);
             }
-            var docs = data.rows;
-            if (form) {
```
SCdF: I dropped this because AFAICT no one uses it

SCdF:
> Just to confirm, have we decided to re-add the original id generate functions that supports check digit?

The version I commited does use the check digit! https://github.com/medic/medic-sentinel/pull/114/files#diff-181a4a985800e3f9449fe4a46d4fe03cR33

In terms of the rest of the complexity in that function, I don't see why we'd keep it:
 - It's more complicated, and implies there is a reason for its complexity (why do you pass in something? Do you need to? Does it matter what you pass in? etc)
 - Makes auto-extension harder (because you either have to change the config format to understand auto-extension, or commit something that is half-broken)
 - And generally doesn't seem necessary/

mandric:
Doh my bad missed that!  But still I don't understand requiring justification for code that is already there and has been there for quite a long time.  It was requested at some point and this ticket has no reason to simplify it in terms of functionality, it's pre-optimization from my pov.  The ticket did not request removing anything so we're trying to remove as little functionality as possible here and still solve the problem.  I see how removing the format config helps to generate/manage an ids pool, but maybe I'm missing something.

SCdF:
I'm confused: I thought this was the decision we came to on our call: you brought up that no one used that app setting, and we took a look and the only use was once to make it larger (which I think you did). We have to remove the app setting so we can auto-extend it, right? Or we need to design an extension to the format to support multiple lengths (i.e. if you think that allowing people to specify a format is important it stands to reason it's important at longer lengths as well), but that seems nuts for something no one uses.

What is it that you're asking for? How would you write this change?

lib/utils.js
mandric: :+1:

mandric:
Sorry that was a bit confusion, I wrote a bit my summary of the call, but it was not very clear.  Probably a good reason to try to summarize a call in the PR so it's clear to both parties, before moving on. 

The setting will be removed from `kanso.json` so it not available in `app_settings` anymore, and can easily be moved to the transitions setting in the future if we need it.  But I wouldn't remove it completely from the underlying library, there is no reason for that.  But `config.get` can be removed and refactored later if we need it...

SCdF:
OK that all makes sense!

Unfortunately, I do utterly disagree with you though, when it comes to deleting code ;-)

I don't believe in keeping complicated code around that no one uses (this PR replaces ~20 lines of code to generate with 2, ignoring the extra parameter validation checks I added to safely support extending the length), and would much rather that we tend to delete things over keeping them around. There is no evidence that people even wanted it in the first place, AFAICT.

So yeah. Not sure how to solve this one? Knife fight? Egg & Spoon race?

mandric:
Oh I was thinking the good 'ol convincing via reason over voice method we've been using with success as of late!

test/unit/transition_utils.js
```diff
@@ -0,0 +1,56 @@
...
+  sinon.stub(utils, 'getRegistrations').callsArgWith(1, null, [{key: 12345}]);
+  var generate = sinon.stub(ids, 'generate');
+  generate.withArgs(5).returns(12345);
+  generate.withArgs(6).returns(123456);
+
+  var doc = {};
+
+  transitionUtils.addUniqueId({}, doc, function(err, result) {
+    if (err) {
+      test.fail('addUniqueId shouldnt error');
```
mandric: Shouldn't these `test.fail` tests be separate unit tests?  They are also repeated.

transitions/utils.js
```diff
@@ -26,5 +32,32 @@ module.exports = {
...
+        db: db,
+        ids: potentialIds
+    }, function(err, registrations) {
+        if (err) {
+            return callback(err);
+        }
+
+        var uniqueIds = _.reject(potentialIds, function(id) {
+          return _.find(registrations, function(registration) {
+            return registration.key === id;
```
mandric: Little worried about matching if our `patient_id` field is a string or a number here...

mandric: Yeah if I look at a production instance they are strings not numbers.

PR #115, opened 2017-03-10
added functional test for short code/ids generation.

mandric:
Wanted to add some kind of benchmark for the `ids` lib because I really had no clue how well it performs.

### DIFF
--- a/test/functional/ids.js
+++ b/test/functional/ids.js
@@ -1,0 +1,33 @@
+var sinon = require('sinon'),
+    testUtils = require('../test_utils'),
+    config = require('../../config'),
+    ids = require('../../lib/ids');
+
+exports.tearDown = function(callback) {
+    testUtils.restore([
+        config.get
+    ]);
+    callback();
+};
+
+exports['generate 1000 unique short ids with less than 100 collisions'] = function(test) {
+    sinon.stub(config, 'get', function() {
+      return '1111';
+    });
+    var LIMIT = 1000,
+        MAX_RETRIES = 100;
+    var stats = {retries: 0, count: 0},
+        data = {};
+    while (stats.count < LIMIT) {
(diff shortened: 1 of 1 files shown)


SCdF:
@mandric The test looks fine (though I'm not sure why you're calling it a performance benchmark, language difference?), but I'm about to delete this generation code right? We're going to replace the whole thing with `Math.random()*(10 ^ digits)` or whatever right? Do you think the test is still relevant? Just want to make sure we're on the same page.

SCdF:
So @mandric I merged this because I wanted to write tests around this area as well. This test actually crashes my node with an oom error. It also takes forever to do so.

I'll look into optimising it, but I think I may end up deleting it in my changes, because:
 - Unit tests should run fast and test features, not slow and test arbitrary performance
 - I shouldn't have to muck around with node heap settings to run a unit test

mandric:
> Unit tests should run fast and test features, not slow and test arbitrary performance

It's not a unit test, that's why it's a functional test in the `tests/functional` directory.

> I shouldn't have to muck around with node heap settings to run a unit test

I didn't need to, not sure what that could be...  also the test won't need mocking when you remove the `config.get` from that function which I don't think belongs there anyway.  Not sure if that helps though.

mandric:
> I'm about to delete this generation code right? We're going to replace the whole thing with Math.random()*(10 ^ digits) or whatever right? Do you think the test is still relevant? Just want to make sure we're on the same page.

I would keep the function because it has a nice feature of a configurable format, I see no value in throwing that away because it could be useful in the future. I would just pass it the format config as an argument and have a default of `1111`.

SCdF:
I'm way ahead of you sorry, perils or remote development. I'm about to push the PR, we can talk over it soon

mandric:
Sure, np!

PR #116, opened 2017-03-14
External ID Support

SCdF:
medic/medic-webapp#3166


transitions/registration.js
```diff
@@ -150,57 +170,36 @@ module.exports = {
...
                 return callback(null, true);
             }
 
             if (doc.fields && doc.fields.patient_id) {
                 // We're attaching this registration to an existing patient, let's
                 // make sure it's valid
-                // NB: if in the future we allow partners to configure the add_patient
-                //     trigger to support passing in their own patient shortcodes
-                //     we're going to have to change this logic:
-                //     https://github.com/medic/medic-webapp/issues/3000
```
SCdF: This is OK, as long as we document that the external ID they pass cannot be mapped to `patient_id` but must be a different name (like `external_id`). Need to document this somewhere.

transitions/registration.js
```diff
@@ -27,6 +27,26 @@ var getRegistrations = function(db, patientId, callback) {
     utils.getRegistrations({ db: db, id: patientId }, callback);
 };
 
+
+var addValidationErrors = function(registrationConfig, doc, errors) {
```
SCdF: Nothing changed, I just extracted it out to make the place where it's called read cleaner.

transitions/utils.js
```diff
@@ -9,29 +9,53 @@ var IDS_TO_GENERATE = 5;
...
+    Adds a "message" and "error" of the configured key to the report. This
+    indicates something went wrong, and the key indicates what went wrong.
+  */
+  addRejectionMessage: function(document, reportConfig, errorKey) {
+    var foundMessage = {
+      doc: document,
+      // TODO: what should this default front be? Popular options:
+      //       'sys.'
+      //       'messages.generic.'
+      message: 'sys.' + errorKey,
```
SCdF: We use `sys.` in one instance I think, and standard uses `messages.generic.`

transitions/registration.js
```diff
@@ -217,21 +216,29 @@ module.exports = {
...
                 }
-                var options = { db: db, audit: audit, doc: doc };
-                if (event.params) {
-                    // params setting get sent as array
+                var options = { db: db, audit: audit, doc: doc, registrationConfig: registrationConfig };
+
+                if (typeof event.params === 'string') {
+                    // params setting can get sent as string to convert into an array
                     options.params = event.params.split(',');
+                } else if (typeof event.params === 'object') {
```
garethbowen: I see you have unit tests for this, but have you tried using an object on an actual instance? In the kanso.json app_settings schema event params are `"type": "string"` which, looking through the code, should make this branch unreachable. It's definitely a good idea to allow JSON as event params but we might need to JSON.parse it somewhere first.

transitions/registration.js
```diff
@@ -333,14 +340,43 @@ module.exports = {
         var doc = options.doc,
             db = db || options.db;
 
-        transitionUtils.addUniqueId(db, doc, callback);
+        if (options.params.patient_id) {
```
garethbowen: I think the param property could be more clearly named. How about `patient_id_field` or something? This would then be consistent with `patient_name_field` at line 378.

transitions/registration.js
```diff
@@ -333,14 +340,43 @@ module.exports = {
...
+            var located = doc.fields[options.params.patient_id];
```
garethbowen: `located` doesn't make much sense to me (feels like it should be a boolean). How about `providedId` or something?

transitions/registration.js
```diff
@@ -333,14 +340,43 @@ module.exports = {
...
+            transitionUtils.isIdUnique(db, located, function(isUnique) {
+                if (isUnique) {
+                    doc.patient_id = located;
+                    callback();
+                } else {
+                    transitionUtils.addRejectionMessage(
+                        doc,
+                        options.registrationConfig,
+                        'provided_patient_id_not_unique');
+                    return callback(null, true);
```
garethbowen: <nitpick>Inconsistent use of return with line 357 - either use return or `else` block for both. It's fine, but when I read it it makes me question whether I really understand the flow, so I have to read it twice.</nitpick>

transitions/registration.js
```diff
@@ -333,14 +340,43 @@ module.exports = {
...
         var doc = options.doc,
             db = options.db,
             audit = options.audit,
             patientShortcode = doc.patient_id,
-            patientNameField = _.first(options.params) || 'patient_name';
+            patientNameField = (
+                Array.isArray(options.params) ?
+                    _.first(options.params) :
+                    options.params.patient_name_field
+            ) || 'patient_name';
```
garethbowen: Ternary mixed with `||` is confusing. Maybe pull this out as a function...

```
var getPatientNameField = function(params) {
  if (params && params.patient_name_field) {
    return params.patient_name_field
  }
  if (Array.isArray(params) && params.length) {
    return params[0];
  }
  return 'patient_name';
};
```

transitions/utils.js
garethbowen: Yeah I'm not a fan of 'sys.'. Go with 'messages.', or 'messages.generic' IMO

transitions/utils.js
```diff
@@ -9,29 +9,53 @@ var IDS_TO_GENERATE = 5;
...
+  addRegistrationNotFoundError: function(document, reportConfig) {
+    module.exports.addRejectionMessage(document, reportConfig, 'registration_not_found');
+  },
+  isIdUnique: function(db, id, callback){
+    utils.getRegistrations({
+        db: db,
+        id: id
+    }, function(err, registrations) {
+        if (err) {
+            callback(err);
```
garethbowen: This is weird - if you get an `err` that will be interpreted as unique because it's truthy. Also it's completely breaking the convention. Instead invoke the callback with two params: `err` and `isUnique`.

transitions/utils.js
SCdF: Cool.

transitions/registration.js
SCdF: That's what real testing is for! :P 

But right OK, that's awful. I might need to re-think this. Do you know how this stuff works? Can I make the type both?

transitions/registration.js
SCdF: Cool, will do.

transitions/registration.js
SCdF: Sorry yeah, that was a left-over from a different strategy, you're right it makes no sense now.

transitions/registration.js
SCdF: rgr that

transitions/utils.js
SCdF: You're right, fixed

transitions/registration.js
garethbowen: I don't think there's a JSON type. There's an object type but that expects the object to be defined not freetext. I've been thinking for a while we should deprecate the comma separated params in favour of JSON array (or object in this case) style, so we could add a second property for json_params (come up with a better name)? Or just take the string and JSON.parse, catch the error, try splitting, give up?

SCdF: Ugh.

SCdF: OK. I cannot think of a clean way of doing this, but maybe that's the Friday in me speaking. Let's not push it. Let's talk Monday. Ideally what I want, is for people to be able to add actual JSON (not a String we all have a secret handshake to agree is JSON). If that has to be a different property, I'd suggest we think of a better name than "param" and then deprecate its usage in favour of this one.

PR #117, opened 2017-03-16
Better optimisation of id generation medic/medic-webapp#3231

SCdF:


transitions/registration.js
```diff
@@ -299,7 +291,8 @@ module.exports = {
         if (!config.messages || !config.messages.length) {
             return callback();
         }
-        getRegistrations(db, patientId, function(err, registrations) {
+        // TODO: can patientId be empty here?
```
garethbowen: Remove TODO - fix the problem or raise an issue.

transitions/registration.js
```diff
@@ -318,10 +311,11 @@ module.exports = {
         });
     },
     assignSchedule: function(options, callback) {
-        getRegistrations(
-            options.db,
-            options.doc.fields && options.doc.fields.patient_id,
-            function(err, registrations) {
+        // TODO: can patientId be empty here?
```
garethbowen: Remove TODO - fix the problem or raise an issue.

transitions/utils.js
```diff
@@ -44,44 +94,37 @@ module.exports = {
...
-        } else {
-            callback(true);
-        }
+    db.medic.view('medic', 'registered_patients', {
+      key: id
+    }, (err, registrations) => {
+      if (err) {
+          callback(err);
+      } else if (registrations.length) {
+          callback(false);
```
garethbowen: Change this to pass `(null, false)`

transitions/utils.js
```diff
@@ -44,44 +94,37 @@ module.exports = {
...
+      } else {
+          callback(true);
```
garethbowen: Change this to pass `(null, true)`

transitions/utils.js
```diff
@@ -44,44 +94,37 @@ module.exports = {
...
-        });
-
-        if (!uniqueIds.length) { // id collision, retry
-            logger.warn('Could not create a unique id of length ' + currentIdLength + ', increasing by one');
-            currentIdLength += 1;
-            module.exports.addUniqueId(db, doc, callback);
-        } else {
-            doc.patient_id = uniqueIds[0];
-            callback();
+        if (!newCacheSize) {
```
garethbowen: I wouldn't bother returning newCacheSize, and just check idCache.size

SCdF:
re: generators, Set already provides one: that is functionally what `.values()` returns you. I did consider just exposing that, but I didn't do it in the end. I think a better idea is to create an actual stateful generator if possible, so that it completely encapsulates and manages refreshing ids from the DB, scaling up etc. I don't know how that would work with the I/O callback stuff.

I'll fix the rest of this PR etc and maybe come back to it.

transitions/utils.js
```diff
@@ -41,44 +38,27 @@ module.exports = {
...
-
-        if (!uniqueIds.length) { // id collision, retry
-            logger.warn('Could not create a unique id of length ' + currentIdLength + ', increasing by one');
-            currentIdLength += 1;
-            module.exports.addUniqueId(db, doc, callback);
-        } else {
-            doc.patient_id = uniqueIds[0];
-            callback();
-        }
+    console.log('addUniqueId');
```
garethbowen: Accidental console.log (plus 2 more below)

transitions/utils.js
```diff
@@ -41,44 +38,27 @@ module.exports = {
...
+    idGenerator.next().value.then(patientId => {
+      console.log('gotPatientId', patientId);
+      doc.patient_id = patientId;
+      callback();
+    }).catch(err => {
+      console.log('poop happened', err);
```
garethbowen: Unacceptable profanity! :hankey:

transitions/utils.js
```diff
@@ -41,44 +38,27 @@ module.exports = {
...
-        if (!uniqueIds.length) { // id collision, retry
-            logger.warn('Could not create a unique id of length ' + currentIdLength + ', increasing by one');
-            currentIdLength += 1;
-            module.exports.addUniqueId(db, doc, callback);
-        } else {
-            doc.patient_id = uniqueIds[0];
-            callback();
-        }
+    console.log('addUniqueId');
+    idGenerator.next().value.then(patientId => {
```
garethbowen: I feel like this is exposing too much of the implementation. Can the generator be a constant in the `ids` module and we just expose a `generate()` method which does the `.next().value` and just returns the promise? That would make this code...

```
ids.generate().then(...).catch(...);
```

Feels cleaner somehow...

.jshintrc
```diff
@@ -38,7 +38,7 @@
     "debug"         : false,     // true: Allow debugger statements e.g. browser breakpoints.
     "eqnull"        : true,      // true: Tolerate use of `== null`
     "es5"           : false,     // true: Allow ES5 syntax (ex: getters and setters)
-    "esnext"        : false,     // true: Allow ES.next (ES6) syntax (ex: `const`)
+    "esnext"        : true,      // true: Allow ES.next (ES6) syntax (ex: `const`)
```
garethbowen: You can replace `es5` and `esnext` with `"esversion"     : 6,` which seems nicer (esnext will sound stupid once es7 is out... `esnextnext`?)

lib/ids.js
```diff
@@ -7,36 +11,114 @@ two numbers results in a different checksum digit, and thus an invalid id.
...
+const generateId = function(length) {
+  if (length && typeof length !== 'number') {
+    throw new Error('generate requires that you pass it a length 0 < x < 14');
+  }
 
-      if (length >= 14) {
-        // TODO: when we support a million billion patients per installation change
-        //       the algorithm to support it ;-)
-        throw new Error('WARNING: id length of ' + length + ' is too long');
+  if (length >= 14) {
```
SCdF: done

lib/ids.js
```diff
@@ -7,36 +11,114 @@ two numbers results in a different checksum digit, and thus an invalid id.
...
+    // TODO: when we support a million billion patients per installation change
+    //       the algorithm to support it ;-)
+    throw new Error('WARNING: id length of ' + length + ' is too long');
+  }
+
+  const randomDigits = String(Math.random() * 10).replace('.','');
+
+  return addCheckDigit(randomDigits.substring(0, length - 1));
```
SCdF: I'll refactor it. I also realised it's more complicated than it needs to be, it should really be `String(Math.random()).replace('0\.', '')`

SCdF: Also, while we're here: how would you generate N random digits in a easier to read fashion?

SCdF: OK I tried to clean it up a little. I honestly don't think it looks any better: at some point you have to read the complexity. Let me know how it goes.

lib/ids.js
```diff
@@ -7,36 +11,114 @@ two numbers results in a different checksum digit, and thus an invalid id.
...
+      currentIdLength = DEFAULT_ID_LENGTH;
+
+  // Developers NB: if you set the cache size too high it will take forever
+  // or potentially be impossible to actually generate enough unique randomly
+  // generated ids.
+  if (MAX_IDS_TO_CACHE * 10 > Math.pow(10, DEFAULT_ID_LENGTH)) {
+    throw new Error('MAX_IDS_TO_CACHE too high compared to DEFAULT_ID_LENGTH');
+  }
+
+  const getNextValue = function(resolve, reject) {
```
SCdF: You're totally right, I hadn't mentally transitioned back to promises properly. Fixed.

transitions/utils.js
SCdF: Whoops… I'm wondering if we want to make jshint explode on travis with console.log (not locally)

SCdF: Same with `.only` in tests

transitions/utils.js
SCdF: 💩 !

transitions/utils.js
SCdF: I get what you're saying. Seems like a bit of a toss up though honestly. In this implementation, `ids` is a stateless module that provides a "factory" (remember that word?) to generate a stateful iterator. The downside being, whomever wants that iterator has to iterate over it. In your version `ids` would hold that state. It's just where you want the state to sit, and thus where the iteration occurs.

I am personally OK with it being where it is. I think the ugliness that you're worried about is actually just inherent to JS generators (I wish `.next()` just got you the value, but I guess they didn't want logic to be confused around whether or not the iterator was done due to falsy. Perhaps they should have `hasNext()` and `next()` for those who need that distinction). If you want to play around with moving it feel free to give it a crack.

.jshintrc
SCdF: I did not know you can do that! I will do so at once!

lib/ids.js
garethbowen: I think it's basically there now. I feel like it should be easier but I'm not sure that it is. I guess it's something people don't do very often...

server.js
```diff
@@ -12,6 +12,16 @@ if (process.env.TEST_ENV) {
...
 
+console.log('Node Version:', process.version);
+var version = process.versions.node.match(/(\d)+\.(\d)+\.(\d)+/)[1];
+if (Number(version[1] <= 4)) {
+  // 5 seems to be where the majority of ES6 was added without flags.
+  // Seems safeist to not allow api to run
+  throw new Error('Node version ' + process.version + ' is not supported');
+}
+if (Number(version[1]) < 6 && Number(version[2]) < 10) {
+  console.error('This node version may not be supported');
```
garethbowen: Make this a warn.

transitions/utils.js
```diff
@@ -41,44 +38,24 @@ module.exports = {
...
-            currentIdLength += 1;
-            module.exports.addUniqueId(db, doc, callback);
-        } else {
-            doc.patient_id = uniqueIds[0];
-            callback();
-        }
+    idGenerator.next().value.then(patientId => {
+      doc.patient_id = patientId;
+      callback();
+    }).catch(err => {
```
garethbowen: Just `.catch(callback)` ?

server.js
garethbowen: Also check with @browndav but I think we're shipping 6.9 on medic-os so probably best to use that.

SCdF: Hmm I think I'm losing my mind: I swear `console.warn` didn't used to exist… it seems to exist in both node and the browser though, so IDK. Must be thinking of something else. I'll change it.

In terms of changing the version, maybe we should line up travis and medic-os then? The wording doesn't come from what medic-os ships, it comes from what we officially test on. So either: we should change travis or we should change medic-os. Thoughts @browndav / @garethbowen ?

SCdF: (actually we have a logger in sentinel, I'll use that as well)

transitions/utils.js
SCdF: 100%, good catch

PR #118, opened 2017-03-22
Don't exit on configuration change

garethbowen:
Reload the config in place rather than forcing an exit. This is a better design and also makes e2e testing much easier.

medic/medic-webapp#3229


config.js
```diff
@@ -1,83 +1,81 @@
...
-            logger.info('Detected translations change - reloading');
-            loadTranslations();
-        }
+const loadTranslations = function() {
+  const options = {
+    startkey: [ 'translations', false ],
+    endkey: [ 'translations', true ],
+    include_docs: true
+  };
+  db.medic.view('medic-client', 'doc_by_type', options, function(err, result) {
```
SCdF: short-form fn def here?

config.js
```diff
@@ -1,83 +1,81 @@
...
+    if (err) {
+      logger.error('Error loading translations - starting up anyway', err);
+      return;
+    }
+    result.rows.forEach(function(row) {
```
SCdF: short-form fn def here?

config.js
```diff
@@ -1,83 +1,81 @@
...
 
-    feed.follow();
-}
+const initFeed = function() {
+  // Use since=now on ddoc listener so we don't replay an old change.
+  const feed = new follow.Feed({ db: process.env.COUCH_URL, since: 'now' });
+  feed.on('change', function(change) {
+    if (change.id === '_design/medic') {
+      logger.info('Reloading configuration');
+      initConfig(function(err) {
```
SCdF: shortform fn def here

config.js
```diff
@@ -1,83 +1,81 @@
...
+        if (err) {
+          console.error('Error loading configuration. Exiting...');
+          process.exit(0);
+        }
+        require('./transitions').loadTransitions();
+      });
+    } else if (change.id.indexOf('messages-') === 0) {
```
SCdF: FYI Node now supports `startsWith`, i.e. `else if (change.startsWith('messages-')) {`

config.js
```diff
@@ -1,83 +1,81 @@
...
+  });
+  feed.follow();
+};
 
-function initConfig(data, callback) {
-    var settings = data.settings;
-    // merge defaults with app settings
-    _.extend(config, settings);
+const initConfig = function(callback) {
+  db.medic.get(SETTINGS_PATH, function(err, data) {
```
SCdF: I'm going to stop mentioning these, but IMO you should go through and change them all :-)

PR #119, opened 2017-03-23
Make it clear that the number being logged is a running total

alxndrsn:


transitions/index.js
```diff
@@ -53,7 +53,7 @@ var processChange = function(change, callback) {
         }
 
         if (processed > 0 && (processed % PROGRESS_REPORT_INTERVAL) === 0) {
-            logger.info('transitions: %d items processed', processed);
+            logger.info('transitions: %d items processed', processed, '(total)');
```
mandric: This still isn't very clear to me unfortunately...

mandric: Also is it a pain to use the printf style format like in the previous log line?

alxndrsn: > This still isn't very clear to me unfortunately...

Any suggestions for improving the message?

> Also is it a pain to use the printf style format like in the previous log line?

Good call - will fix.

mandric: To me it looks like it would print:

alxndrsn: Improvement(?) at https://github.com/medic/medic-sentinel/pull/126

PR #122, opened 2017-03-25
Add bikram sambat date format. medic/medic-webapp#3219

estellecomment:


estellecomment:
Tests are probably breaking because it needs an update to `bikram-sambat` package.

lib/template.js
```diff
@@ -1,24 +1,34 @@
...
+var format = function(date, formatString) {
   if (!isNaN(date)) {
     date = parseInt(date, 10);
   }
-  return moment(date).format(config.get(key));
+  return moment(date).format(formatString);
 };
 
 var formatters = {
+  bikram_sambat_date: function() {
```
alxndrsn: I wonder if this can be made more concise.  Good to have `_date` in there for people who don't know what `bikram_sambat` means; but would e.g. `bikram_date` make sense to Nepali tech leads/project managers?

alxndrsn: Or perhaps `bs_date`, `np_date`?

PR #124, opened 2017-03-27
Correct trigger param parsing

SCdF:
If `params` doesn't parse right we were just dropping the error (yay callbacks!). I've refactored the code so that the error can be passed back through the main callback correctly.

First commit is just the tests, second is the code what fixes it.

image

### DIFF
--- a/transitions/registration.js
+++ b/transitions/registration.js
@@ -234,17 +234,29 @@ module.exports = {
                     if (!event.params) {
                         options.params = {};
                     } else {
-                        // We currently support JSON in a string:
-                        // "{\"foo\": \"bar\"}"
-                        // And comma delimted strings:
-                        // "foo,bar" or just "foo"
-                        try {
-                            options.params = JSON.parse(event.params);
-                        } catch (e) {
+                        if (event.params instanceof Object) {
+                            // We support raw JSON even if we can't specify that
+                            // correctly in kanso.json
+                            options.params = event.params;
+                        } else {
                             try {
-                                options.params = event.params.split(',');
-                            } catch (e) {
-                                return cb(new Error('event params cannot be parsed: must be either a comma-delimited String, or JSON *inside* a String'));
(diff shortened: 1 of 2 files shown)


transitions/registration.js
```diff
@@ -218,36 +218,44 @@ module.exports = {
...
+                        // "{\"foo\": \"bar\"}"
+                        // And comma delimted strings:
+                        // "foo,bar" or just "foo"
+                        try {
+                            options.params = JSON.parse(event.params);
+                        } catch (e) {
+                            try {
+                                options.params = event.params.split(',');
+                            } catch (e) {
+                                return cb(new Error('event params cannot be parsed: must be either a comma-delimited String, or JSON *inside* a String'));
```
garethbowen: "JSON inside a string" doesn't make sense to me. Can you say "serialized JSON" or is that too technical?

garethbowen: Also, maybe include both `e`s in the error somewhere - they might be useful.

SCdF: Yeah that was the least technical I could think to make that error on one line. However, whatever, let's make it really detailed with multiple lines and examples! It's the multi-line-string future!

SCdF:
@garethbowen take another look please. I changed my approach a reasonable amount. The previous approach wouldn't actually detect *bad* JSON because it would just fall back to splitting a string.

So if you passed: `"{\"foo\": \"bar\""` (missing the closing `}`) it wouldn't actually error, because it would `JSON.parse()` that, fail, then `.split(',')` it, which would result in params equalling the entire string. Since params aren't required necessarily it would just happy process the doc and move on with its life.

This attempt does two things differently:
 - It supports mapping raw JSON. I'm not convinced our entire app will support this properly due to kanso.json expecting a string here, so it's still not our official method, but it's there anyway
 - We try to detect if your string is supposed to be JSON and just parse it that way, instead of falling back on string.split

PR #125, opened 2017-03-28
Adds QuickCheck style test with jsverify

SCdF:


test/unit/ids.js
```diff
@@ -16,26 +17,31 @@ const mockDb = (idFilterLogicFn) => {
...
 
-    test.equal(ids._generate(5), '00000');
-    test.done();
+  test.equal(ids._generate(5), '00000');
+  test.done();
+};
+
+exports['ids are "always" the length they should be'] = test => {
+  test.ok(
+    jsc.checkForall(jsc.integer(5, 13), i => ids._generate(i).length === i));
```
mandric: I don't understand why you would need a different library to do this, but maybe you can enlighten me?

SCdF: I agree it's hard to tell from one line!

In terms of whether we need a library, I'll say two things:
 - If you're asking if we should be doing property based testing at all, then the answer is I don't know. I think it could be handy in some very specific cases, specifically around automated bounds testing of more "algorithm-y" things
 - If you're asking if we need a library in general because dependencies are bad, then IDK, I mean we don't really need mocha / nodeunit / jasmine either honestly, they just execute JS and catch exceptions. 

It may be worthwhile reading a bit about QuickCheck and friends, and they are bit weird and do a bit more than may be obvious.

FWIW I may be adding some around dates in the near future that will probably look more impressive, and will probably be a more useful evaluation of its worth.

mandric: Ok, so I don't understand what it's doing then because I'm unfamiliar with QuickCheck. It looks to me like you're testing the same thing as the `generates an id of the given length` test.  If you want to enlighten me re QuickCheck that's cool, otherwise I'll read up on it soon, thanks!

SCdF: @mandric so that is definitely *half* true!

Basically, the difference is:
 - In the normal test, I am specifically knowledgable of a potential bug in an id generation implementation where if the id starts with 0 and you make the mistake to treat the id as a number, you will loose the 0s on the front. I am then **hard-coding my testing of this known edge case**
 - In the jsverify / quickcheck / property based check / generative test (first two are libraries, second two are names for basically the same thing, for your later googling) I am **declaring a property that this algorithm has**. It is then up to the library to exercise and properly test that library. This test should catch the specific edge case I know about, but it could catch others as well.

Here is a better example: let's say I'm writing a sort algorithm. There are two obvious general properties of any sort algorithm:
 - `items.length === sorted(items).length`
 - `items[i] <= items[i++]`

There may be more, either general, or specific to your case.

You can test those a few ways:
 - Hard code some examples and presume they cover the edge cases
 - Have a test that randomly generates N collections of items (presumably also at random lengths) and then tests that each one has those properties
 - Use a QuickCheck variant to declare those properties and let the library take care of the testing

In theory each of these ideas is better and more encompassing than the previous. Good implementations of QuickCheck should fail *consistently* (as opposed to just be a flaky test), and should also attempt to narrow down the failures into the most concise (read: smallest data structure) set of inputs that cause the failure.

Again, I'm not 100% sold but I'm interested in giving it a try, and again it's probably not something you can apply everywhere, and more something that is a specific tool for a specific purpose.

mandric: Hey thanks for this explanation, unfortunately I still don't understand.  I probably need one more example or something.  Anyway didn't mean for you to spend too much time on it.  Basically I'm completely lost in the paragraph below, probably because I have no experience with QuickCheck.

"In the jsverify / quickcheck / property based check / generative test (first two are libraries, second two are names for basically the same thing, for your later googling) I am declaring a property that this algorithm has. It is then up to the library to exercise and properly test that library. This test should catch the specific edge case I know about, but it could catch others as well."

PR #127, opened 2017-04-06
medic/medic-webapp#2180 WIP

SCdF:
WIP. Plz don't merge kthnxbye


README.md
```diff
@@ -125,10 +125,14 @@ Transitions obey the following rules:
...
-  copying is discouraged.
+* has an `onChange(change, db, audit, callback)` function that makes changes to
+  the `change.doc` reference (copying is discouraged). `db` and `audit` are
+  handles to let you query those DBs. More about `callback` below.
+
+* takes responsibility for saving the document and re-attaching the newly saved
+  document (with new seq etc) at `change.doc`
   
-* does not have side effects outside of the document passed in.  This might be
+* does not have side effects outside of `change.doc`.  This might be
```
garethbowen: This isn't true any more is it? We do things like create patients. Maybe say that "ideally" it would have no side effects?

transitions/add_info_document.js
```diff
@@ -0,0 +1,77 @@
...
+                   result.doc.history.find(function(el) {
+        return el.action === 'create';
+      });
+
+      if (create) {
+        callback(null, {
+          _id: infoDocId(docId),
+          type: 'info',
+          doc_id: docId,
+          initial_replication_date: create.timestamp
```
garethbowen: Clever!

transitions/add_info_document.js
```diff
@@ -0,0 +1,77 @@
...
+  db.medic.get(infoDocId(docId), function(err, body) {
+    if (err && err.statusCode !== 404) {
+      callback(err);
+    } else {
+      callback(null, body);
+    }
+  });
+};
+
+var generateInfoDocFromAuditTrail = function(audit, docId, callback) {
```
garethbowen: I think it'd be cleaner if this function just got the created date, either create.timestamp or 'unknown'. There's a bit of duplication with filling in the other doc fields.

transitions/index.js
```diff
@@ -21,6 +21,7 @@ var _ = require('underscore'),
  */
 var AVAILABLE_TRANSITIONS = [
     'accept_patient_reports',
+    'add_info_document',
```
garethbowen: I don't think this transition should be optional. Although you can live without it for now I think we'll soon start depending on the doc, so let's just hardcode it for everyone.

transitions/index.js
```diff
@@ -73,24 +103,34 @@ const processChange = (change, callback) => {
  * Load transitions using `require` based on what is in AVAILABLE_TRANSITIONS
  * constant and what is enabled in the `transitions` property in the settings
  * data.  Log warnings on failure.
+ *
+ * Tests can diabled autoEnableSystemTransitions for more control.
  */
-const loadTransitions = () => {
+const loadTransitions = (autoEnableSystemTransitions = true) => {
```
garethbowen: Oooh fancy!

transitions/index.js
```diff
@@ -73,24 +103,34 @@ const processChange = (change, callback) => {
...
+  AVAILABLE_TRANSITIONS.forEach(transition => {
+    const conf = transitionsConfig[transition];
+
+    const system =
+      autoEnableSystemTransitions && SYSTEM_TRANSITIONS.includes(transition);
+
+    const disabled = conf && conf.disable;
+
+    if ((system && disabled) || (!system && !conf || disabled)) {
+      return logger.warn(`transition ${transition} is disabled`);
```
garethbowen: Is there a typo here? It looks like you're allowing disabling system transitions, which I don't think is what you intended? I don't think we should allow disabling.

This logic is really confusing but I think this boils down to `if(disabled)` which is why I think there's a typo. If it's not a typo then this needs to be made much easier to read.

transitions/index.js
garethbowen: Nevermind - you've covered it in a later commit!

README.md
SCdF: Yeah, these documentation changes are from feb, and re-reading them and docs around them, I think they are pretty wrong. For example the transition doesn't need to save the doc, that was a misunderstanding of mine. I'll re-read and re-write.

transitions/index.js
SCdF: Man… this `if` statement honestly took me an hour. No joke. I mean it worked in 5 minutes, but I kept on redoing it to try to make it clearer, because evidently it's a really challenging thing to make readable.

But yes: it is intentional that you can disable system transitions. There is documentation against the `SYSTEM_TRANSITONS` const to that effect.

At least, that was the intention when I first wrote it. Standing back from it now I'm not sure if that's still a relevant thing to do, see my comment on the ticket. I'm going to leave this for now, we can always change it later.

transitions/add_info_document.js
SCdF: I pulled out the creation of the base info doc. I presume that is what you mean, but I'm not completely clear.

PR #128, opened 2017-04-21
Using sinon.sandbox for clearer, safer tests

SCdF:
Replacing our pattern of passing a list of things that have maybe been
stubbed to a utility that restores them, and replacing it with sinon
sandboxes, the correct way to deal with this kind of thing.

This is better because it's fool-proof, you do not have to know what
things may or may not have been stubbed.

See: http://sinonjs.org/releases/v2.1.0/sandbox/


test/unit/default_responses.js
```diff
@@ -111,7 +95,7 @@ exports['pass filter when message is not from gateway'] = function(test) {
 };
 
 exports['do nothing if reported date is not after config start date'] = function(test) {
-    testUtils.restore([transition._isReportedAfterStartDate]);
+    transition._isReportedAfterStartDate.restore();
```
alxndrsn: Is there a reason not to just blindly restore the whole lot here?

SCdF: For this exact test it's because you're re-stubbing the function just below. For the others though, I'm actually not that sure. Deleting the individual restores breaks. I'll look into it.

SCdF: Right, and for the others it's because they are testing that actual function so they need the stub dropped.

PR #132, opened 2017-05-15
Adds support for linked contacts

garethbowen:
Minify the docs to the linked format before saving to avoid saving
useless data. Hydrate the docs before using them so parents and
contacts are available and up to date.

medic/medic-webapp#2635


transitions/index.js
```diff
@@ -75,7 +76,7 @@ const processChange = (change, callback) => {
   if (!change) {
     return callback();
   }
-  db.medic.get(change.id, (err, doc) => {
+  utils.hydrateDoc(db, change.id, (err, doc) => {
```
mandric: Wondering if it would be cleaner if hydrateDoc took an id and a list of docs (contacts) and keep the db query out of that function?

transitions/utils.js
```diff
@@ -56,5 +56,18 @@ module.exports = {
       doc.patient_id = patientId;
       callback();
     }).catch(callback);
+  },
+  extractLineage: contact => {
```
mandric: Is it strange that hydrateDoc and extractLineage are not in the same module?

lib/utils.js
```diff
@@ -82,411 +82,450 @@ var truncateMessage = function(parts, max) {
...
-                if (!results.rows.length) {
-                    return callback();
-                }
-
-                if (results.rows.length > 1) {
-                    console.warn('More than one patient person document for shortcode ' + patientShortcodeId);
-                }
-
-                return callback(null, results.rows[0].id);
+        contacts.rows.forEach(contactRow => {
```
estellecomment: make a `hydrateLineageRows` function, that'll make it much easier to read because it's named. (or use a comment, but you know, comments are documentation)

lib/utils.js
```diff
@@ -82,411 +82,450 @@ var truncateMessage = function(parts, max) {
...
-    },
-    /*
-     * Used to avoid infinite loops of auto-reply messages between gateway and
-     * itself.
-     */
-    _isMessageFromGateway: function(from) {
-        var gw = config.get('gateway_number');
-        if (typeof gw === 'string' && typeof from === 'string') {
-            return phoneUtil.isNumberMatch(gw, from) >= 3;
+        const resultRow = lineage.rows.shift();
```
estellecomment: and `buildDocFromHydratedLineage` or something

lib/utils.js
```diff
@@ -82,411 +82,450 @@ var truncateMessage = function(parts, max) {
...
+   */
+  _isMessageFromGateway: from => {
+    const gw = config.get('gateway_number');
+    if (typeof gw === 'string' && typeof from === 'string') {
+      return phoneUtil.isNumberMatch(gw, from) >= 3;
+    }
+    return false;
+  },
+  hydrateDoc: (db, id, callback) => {
+    db.medic.view('medic-client', 'docs_by_id_lineage', {
```
estellecomment: Also mentioned in other PR : `docs_by_id_with_lineage` is clearer

schedule/reminders.js
```diff
@@ -115,18 +115,19 @@ module.exports = {
         var db = options.db;
 
         // gets all clinics
-        db.medic.view('medic', 'clinic_by_phone', {
+        db.medic.view('medic-client', 'doc_by_type', {
```
estellecomment: that's just unrelated! Lucky for you that @alxndrsn is not reviewing this, cos it would NOT FLY 😛

test/unit/update_sent_by.js
```diff
@@ -4,96 +4,36 @@ var sinon = require('sinon').sandbox.create(),
...
-
-    transition.onMatch({
-        doc: doc
-    }, fakedb, fakeaudit, function(err, changed) {
-        test.ok(changed);
-        test.equal(doc.sent_by, 'Clinic');
-        test.ok(dbView.calledOnce);
-        test.done();
-    });
+exports['updates sent_by to clinic name if contact name'] = function(test) {
```
estellecomment: not quite following why some tests are gone and another one appears, but assuming you know why

transitions/index.js
estellecomment: True, `fetchHydratedDoc` would be a less sneaky name

transitions/utils.js
estellecomment: yes it is! 
They should also be called with matching names, e.g. `hydrateDoc` and `dehydrateDoc`

test/unit/update_sent_by.js
garethbowen: The implementation changed so those tests were no longer relevant.

schedule/reminders.js
garethbowen: It's actually related - clinic_by_phone was broken due to the linked contacts changes IIRC so I either had to update it or replace it.

transitions/utils.js
garethbowen: Yeah I should pull these two out into a LineageUtils or something.

lib/utils.js
garethbowen: Hmm maybe. We use words after "by" to indicate it's part of the key, and the lineage depth is part of the key. In the view definition it _looks_ like it's the value, but what you get back is just a list of docs. Let's talk about this one...

lib/utils.js
garethbowen: :+1:

lib/utils.js
garethbowen: :+1:

transitions/index.js
garethbowen: :+1:

PR #133, opened 2017-05-30
3505 add id length doc

SCdF:
plz don't squash.

### DIFF
--- a/lib/ids.js
+++ b/lib/ids.js
@@ -1,9 +1,10 @@
 const _ = require('underscore'),
       logger = require('./logger');
 
-const INITIAL_ID_LENGTH = 5,
-      ID_LENGTH_DOC_ID = 'shortcode-id-length',
+const ID_LENGTH_DOC_ID = 'shortcode-id-length',
       ID_LENGTH_PARAM = 'current_length',
+      MIN_ID_LENGTH = 5,
+      INITIAL_ID_LENGTH = MIN_ID_LENGTH,
       MAX_ID_LENGTH = 13,
       MAX_IDS_TO_CACHE = 100;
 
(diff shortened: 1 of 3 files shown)


lib/ids.js
```diff
@@ -1,7 +1,10 @@
 const _ = require('underscore'),
       logger = require('./logger');
 
-const DEFAULT_ID_LENGTH = 5,
+const ID_LENGTH_DOC_ID = 'shortcode-id-length',
```
mandric: If sentinel is already using a meta doc in other parts of the code we might also considering using it, or at least a similar naming schema so it's easy to find them all.  Then we cat fetch them all  with `{startkey: 'meta-sentinel-', endkey: 'meta-sentinel-\uffff'}` etc.

SCdF: The only one I'm aware of is `sentinel-meta-data`, which is a bit unfortunate for what you're suggesting. FWIW I'd like to actually move `sentinel-meta-data`, `shortcode-id-length` as well as transition information (currently stored in docs) out into a separate sentinel DB. This would help that greatly I think (as it would also be an opportunity to rename `sentinel-meta-data` to something else).

lib/ids.js
```diff
@@ -33,13 +36,21 @@ const randomDigits = (length) =>
 
 const generateId = (length) => {
   if (length && typeof length !== 'number') {
-    throw new Error('generate requires that you pass it a length 0 < x < 14');
+    throw new Error('generateId requires that you pass it a length 5 <= x <= 13');
```
SCdF: Nope, I'll change it.

lib/ids.js
```diff
@@ -48,11 +59,41 @@ const generateId = (length) => {
...
+    db.medic.put(idLengthDoc, err => {
+      // We're OK with a 409, because we're going to presume this is happening
+      // because a human edited it to suite their needs, and their write is more
+      // important than ours.
+      if (err && err.statusCode !== 409) {
+        reject(err);
+      } else {
+        resolve();
+      }}));
+
```
SCdF: They currently are being essentially ignored yeah. I'll add a `console.log` about it.

lib/ids.js
```diff
@@ -48,11 +59,46 @@ const generateId = (length) => {
...
+const putIdLengthDoc = (db, idLengthDoc) =>
+  new Promise((resolve, reject) =>
+    db.medic.put(idLengthDoc, err => {
+      if (err) {
+        // We're OK with a 409, because we're going to presume this is happening
+        // because a human edited it to suite their needs, and their write is more
+        // important than ours.
+        if (err.statusCode === 409) {
+          console.log(`409 while trying to store ${idLengthDoc}. If someone edited this document it is expected.`);
```
mandric: Any reason you don't use `logger.warn` here?

SCdF: Lol, lord no. Sorry. Too many code bases and different conventions. I'll change it.

PR #134, opened 2017-06-09
Accept patient_id as well as fields.patient_id. medic/medic-webapp#3536

estellecomment:
medic/medic-webapp#3536


garethbowen:
The root cause was found to be a migration that failed. This is no longer needed.

PR #135, opened 2017-06-15
Bug fix - check for undefined

henokgetachew:
This has been happening a lot in the sentinel logs of christianaidisiolo. The issue is here: medic/medic-webapp/issues/3556

### DIFF
--- a/transitions/update_clinics.js
+++ b/transitions/update_clinics.js
@@ -36,7 +36,7 @@ var getContact = function(db, doc, callback) {
             if (err) {
                 return callback(err);
             }
-            if (!data.rows.length) {
+            if (typeof(data) == 'undefined' || !data.rows.length) {
                 return callback();
             }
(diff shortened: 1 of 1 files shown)


transitions/update_clinics.js
```diff
@@ -36,7 +36,7 @@ var getContact = function(db, doc, callback) {
             if (err) {
                 return callback(err);
             }
-            if (!data.rows.length) {
+            if (typeof(data) === 'undefined' || !data.rows.length) {
```
SCdF: A shorter way of doing this is just checking `data` directly, ie `if (!data || !data.rows.length)`.

Regardless though, this is actually a problem with how nano reports some kinds of failure. The correct solution is not to ignore it like there are no results, but to pass an error to the callback.

Ideally we would actually fix Nano, and I'll look into that, but the general rule is that if there is no error but **also** no result, something has gone wrong.

PR #136, opened 2017-07-04
3600 hydrating reports

estellecomment:
medic/medic-webapp#3600

### DIFF
--- a/lib/lineage.js
+++ b/lib/lineage.js
@@ -5,7 +5,7 @@ const bindContacts = (lineage, contacts) => {
     const stub = lineage.find(lineageDoc => {
       const id = lineageDoc.contact &&
                  lineageDoc.contact._id;
-      return id === (contactDoc._id);
+      return id === contactDoc._id;
     });
     if (stub) {
       stub.contact = contactDoc;
@@ -69,12 +69,15 @@ module.exports = {
     if (!doc) {
       return;
     }
-    doc.parent = minifyContact(doc.parent);
+    if (doc.parent) {
+      doc.parent = minifyContact(doc.parent);
+    }
     if (doc.contact && doc.contact._id) {
-      doc.contact = {
-        _id: doc.contact._id,
-        parent: minifyContact(doc.contact.parent)
-      };
+      const miniContact = { _id: doc.contact._id };
+      if (doc.contact.parent) {
+        miniContact.parent = minifyContact(doc.contact.parent);
+      }
+      doc.contact = miniContact;
     }
   }
 };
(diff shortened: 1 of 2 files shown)


lib/lineage.js
```diff
@@ -1,24 +1,26 @@
...
-    const stub = lineage.rows.find(lineageRow => {
-      const id = lineageRow.doc &&
-                 lineageRow.doc.contact &&
-                 lineageRow.doc.contact._id;
-      return id === (contactRow.doc && contactRow.doc._id);
+  contacts.forEach(contactDoc => {
+    const stub = lineage.find(lineageDoc => {
+      const id = lineageDoc.contact &&
+                 lineageDoc.contact._id;
+      return id === (contactDoc._id);
```
garethbowen: Remove parens - not needed due to simpler statement.

test/unit/lineage.js
```diff
@@ -90,6 +90,92 @@ exports['hydrateDoc attaches the lineage'] = test => {
...
+  fetch.callsFake((options, callback) => {
+    const contactDocs = options.keys.map(id => {
+      switch (id) {
+        case givenContact._id:
+          return givenContact;
+        case parentContact._id:
+          return parentContact;
+        case grandparentContact._id:
+          return grandparentContact;
+      }
```
garethbowen: A find may be cleaner here:

```js
return [ givenContact, parentContact, grandparentContact ].find(contact => contact._id === id);
```

test/unit/lineage.js
```diff
@@ -234,3 +320,137 @@ exports['minify minifies the contact and lineage'] = test => {
...
+    });
+    callback(null, { rows: contactDocs.map(doc => { return {doc: doc}; }) });
+  });
+
+  lineage.fetchHydratedDoc(docId, (err, actual) => {
+    test.equals(err, null);
+    lineage.minify(actual);
+    test.deepEqual(given, actual);
```
garethbowen: I'm not convinced this is testing correctly because minify modifies the given object. I _think_ it's ok because `fetchHydratedDoc` is returning a new object. It's safer however to clone `given` (eg: `JSON.parse(JSON.stringify(given))`) before using it to prove that it's actually working and not just referring to the same memory location and therefore will always be deepEqual...

test/unit/lineage.js
```diff
@@ -234,3 +320,137 @@ exports['minify minifies the contact and lineage'] = test => {
...
+  ] });
+  sinon.stub(db.medic, 'fetch').callsArgWith(1, null, { rows: [
+    { doc: givenContact },
+    { doc: parentContact },
+    { doc: grandparentContact }
+  ] });
+  lineage.fetchHydratedDoc(docId, (err, actual) => {
+    test.equals(err, null);
+    lineage.minify(actual);
+    test.deepEqual(given, actual);
```
garethbowen: As above.

PR #137, opened 2017-07-04
Expose person to message formatter medic/medic-webapp#3419

SCdF:
Allow registration messages to use data from the person contact, e.g.
person.name. This is better than using data from previous registrations
as patients who are created through the UI may not have the same
previous registrations as users created through SMS. However, all
patients have person contacts.

medic/medic-webapp#3419

### DIFF
--- a/test/unit/registration.js
+++ b/test/unit/registration.js
@@ -1,6 +1,7 @@
 var sinon = require('sinon').sandbox.create(),
     transition = require('../../transitions/registration'),
     transitionUtils = require('../../transitions/utils'),
+    messages = require('../../lib/messages'),
     utils = require('../../lib/utils'),
     schedules = require('../../lib/schedules'),
     config = require('../../config');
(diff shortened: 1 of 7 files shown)


test/unit/registration.js
```diff
@@ -589,3 +590,48 @@ exports['trigger param configuration parse failure for invalid JSON propagates t
         test.done();
     });
 };
+
+exports['addMessage prepops and passes the right information to messages.addMessage'] = (test) => {
+    const testPhone = '1234',
+          testMessage = 'A Test Message',
+          testRegistration = 'some registrations',
+          testPerson = 'a patient contact';
```
garethbowen: Change this to an object with a name property so you can test the `patient_name` alias works as expected.

transitions/registration.js
```diff
@@ -329,25 +329,30 @@ module.exports = {
...
+              extra = {next_msg: schedules.getNextTimes(doc, now)},
+              patientId = doc.fields && doc.fields.patient_id;
         if (!config.messages || !config.messages.length) {
             return callback();
         }
-        getRegistrations(db, patientId, function(err, registrations) {
+        async.parallel({
+            registrations: _.partial(getRegistrations, db, patientId),
+            person: _.partial(utils.getPatientContact, db, patientId)
+        }, (err, {registrations, person}) => {
```
garethbowen: So awesome!

test/unit/registration.js
SCdF: This test mocks out `messages.getMessage` so that call is never exercised. I've just added a couple of unit tests in `test/unit/messages.js`, fresh off of my dome, so check those out

PR #138, opened 2017-07-05
3416 multi form alerts

estellecomment:
medic/medic-webapp#3416

# Review checklist

- [X] Readable: Concise, well named, follows the [style guide](https://github.com/medic/medic-docs/blob/master/development/style-guide.md), documented if necessary.
- [X] Documented: Announced in Changes.md in plain English : . Configuration and user documentation on [medic-docs](https://github.com/medic/medic-docs/) 
PR : https://github.com/medic/medic-webapp/pull/3609
PR : https://github.com/medic/medic-docs/pull/26

- [X] Tested: Unit and/or e2e where appropriate
- ~~Internationalised: All user facing text~~
- ~~Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in Changes.md.~~

### DIFF
--- a/test/unit/transitions/multi_form_alerts.js
+++ b/test/unit/transitions/multi_form_alerts.js
@@ -26,13 +26,11 @@ exports.setUp = callback => {
 };
 
 const stubFetchHydratedDocs = () => {
-  sinon.stub(lineage, 'fetchHydratedDocPromise').callsFake(id => {
-    if (id === doc._id) { return Promise.resolve(doc); }
-    if (id === reports[0]._id) { return Promise.resolve(hydratedReports[0]); }
-    if (id === reports[1]._id) { return Promise.resolve(hydratedReports[1]); }
-  });
+  sinon.stub(lineage, 'fetchHydratedDocPromise')
+    .withArgs(doc._id).returns(Promise.resolve(doc))
+    .withArgs(reports[0]._id).returns(Promise.resolve(hydratedReports[0]))
+    .withArgs(reports[1]._id).returns(Promise.resolve(hydratedReports[1]));
 };
-
 
 // doc is hydrated before being passed to the transition.
 const doc = {
@@ -125,7 +123,7 @@ exports['validates config : timeWindowInDays'] = test => {
 exports['fetches reports within time window'] = test => {
   sinon.stub(config, 'get').returns([alert]);
   sinon.stub(utils, 'getReportsWithinTimeWindow').returns(Promise.resolve(reports));
-
+  stubFetchHydratedDocs();
   transition.onMatch({ doc: doc }, undefined, undefined, () => {
     test.equals(utils.getReportsWithinTimeWindow.callCount, 1);
     test.equals(utils.getReportsWithinTimeWindow.args[0][0], 12344);
--- a/test/unit/transitions/multi_form_alerts.js
+++ b/test/unit/transitions/multi_form_alerts.js
@@ -45,8 +45,8 @@ const doc = {
 };
 
 const reports = [
-  { _id: 'docA', form: 'A' },
-  { _id: 'docB', form: 'B'}
+  { _id: 'docA', form: 'A', contact: { _id: 'contactA' } },
+  { _id: 'docB', form: 'B', contact: { _id: 'contactB' } },
 ];
 
 const hydratedReports = [
@@ -235,13 +235,16 @@ exports['adds multiple messages when multiple recipients are evaled'] = test => 
     [_.defaults({ recipients: [recipient]} , alert)]);
 
   transition.onMatch({ doc: doc }, undefined, undefined, (err, docNeedsSaving) => {
-    test.equals(messages.addMessage.getCalls().length, 2);
+    test.equals(messages.addMessage.getCalls().length, 3); // 3 counted reports, one phone number each.
+    const actualPhones = messages.addMessage.getCalls().map(call => call.args[0].phone);
+    const expectedPhones = [doc.contact.phone, hydratedReports[0].contact.phone, hydratedReports[1].contact.phone];
+    test.deepEqual(actualPhones, expectedPhones);
+
     test.equals(messages.addError.getCalls().length, 0);
 
     test.ok(docNeedsSaving);
     test.done();
   });
-  test.done();
 };
 
 exports['adds message when recipient cannot be evaled'] = test => {
@@ -302,6 +305,22 @@ exports['adds multiple messages when mutiple recipients'] = test => {
     assertMessage(test, messages.addMessage.getCall(3), hydratedReports[1].contact.phone, alert.message);
 
     test.equals(messages.addMessage.getCalls().length, 4);
+
+    test.done();
+  });
+};
+
+exports['dedups message recipients'] = test => {
+  var alertMultipleRecipients = _.defaults({
+    recipients: ['countedReports[0].contact.phone', 'countedReports[0].contact.phone']},
+    alert);
+  sinon.stub(config, 'get').returns([alertMultipleRecipients]);
+
+  transition.onMatch({ doc: doc }, undefined, undefined, () => {
+    test.equals(messages.addError.getCalls().length, 0);
+
+    test.equals(messages.addMessage.getCalls().length, 1); // 2 recipients specified, deduped to 1.
+    test.equals(messages.addMessage.getCall(0).args[0].phone, doc.contact.phone);
 
     test.done();
   });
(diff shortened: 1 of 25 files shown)


test/unit/transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,384 @@
...
+  callback();
+};
+
+exports.setUp = callback => {
+  sinon.stub(utils, 'getReportsWithinTimeWindow').returns(Promise.resolve(reports));
+  sinon.stub(lineage, 'fetchHydratedDocPromise').callsFake(id => {
+    if (id === doc._id) { return Promise.resolve(doc); }
+    if (id === reports[0]._id) { return Promise.resolve(hydratedReports[0]); }
+    if (id === reports[1]._id) { return Promise.resolve(hydratedReports[1]); }
+  });
```
garethbowen: I don't like this pattern because it separates the mocking from the test, so when I come back and look at the test later I can never work out why it's working or not. The best way to do it is having something like the following in each test...

```js
sinon.stub(lineage, 'fetchHydratedDocPromise').withArgs(doc._id).returns(Promise.resolve(doc));
```

This makes it really clear how each unit is working.

Also by doing all the mocking in the `setUp` some tests may be passing incorrectly because you don't expect them to be calling this function, or calling it with these params, so you may miss an error.

And finally it means you can assign it to a variable and check things like the number of times it was called, which you can't do if you do it in the setUp and then throw it away.

test/unit/transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,384 @@
...
+exports['filter validation hasRun'] = test => {
+  test.equal(transition.filter({
+    form: 'x',
+    type: 'data_record',
+    transitions : { multi_form_alerts: 'hi' }
+  }), false);
+  test.done();
+};
+
+const testValidate = (test, alert) => {
```
garethbowen: Maybe rename this to `testInvalidConfiguration` (or something) because it's hardcoded to only test validation failures.

test/unit/transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,384 @@
...
+
+  transition.onMatch({ doc: doc }, undefined, undefined, (err, docNeedsSaving) => {
+    test.equals(messages.addError.getCalls().length, 0);
+    test.equals(messages.addMessage.getCalls().length, 0);
+    test.ok(!docNeedsSaving);
+    test.done();
+  });
+};
+
+const assertMessage = (test, call, recipient, message) => {
```
garethbowen: Modify this to take `messageArgs` not `call` so it doesn't have to know that they came from sinon.

test/unit/transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,384 @@
...
+    test.ok(!err);
+    test.ok(docNeedsSaving);
+    test.done();
+  });
+};
+
+exports['adds message when recipient is evaled'] = test => {
+  const recipient = 'countedReports[0].contact.phone';
+  sinon.stub(config, 'get').returns(
+    [_.defaults({ recipients: [recipient]} , alert)]);
```
garethbowen: You use _.defaults a lot and it's quite hard to read. How about resetting `alert` to the default in the setup, and then just saying `alert.recipients = [recipient]` in the test? That is much more readable to me...

transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,219 @@
...
+      messages = require('../lib/messages'),
+      utils = require('../lib/utils');
+
+const hasRun = (doc) => Boolean(
+  doc &&
+  doc.transitions &&
+  doc.transitions.multi_form_alerts
+);
+
+const getAlerts = () => config.get('multi_form_alerts');
```
garethbowen: Alerts is a bit overloaded - maybe this should be `getAlertConfig`?

transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,219 @@
...
+  const fetchedDocsPromise = docs.reduce(function(promise, doc) {
+    return promise.then((fetchedDocs) => {
+      return lineage.fetchHydratedDocPromise(doc._id)
+        .then(fetchedDoc => {
+          fetchedDocs.push(fetchedDoc);
+          return fetchedDocs;
+        });
+    });
+  }, Promise.resolve([]));
+  return fetchedDocsPromise;
```
garethbowen: I don't think this will perform well if someone has lots of reports. For `n` reports you're making `n` requests to the database. Instead you should write a util to collect up all the parent IDs, unique them, request them all from the DB in one go, and then bind them back into the reports.

transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,219 @@
...
+const countReports = (reports, isReportCountedString) => {
+  return reports.filter((report) => {
+    const context = { report: report, latestReport: reports[0]};
+    try {
+      return vm.runInNewContext('(' + isReportCountedString + ')(report, latestReport)', context);
+    } catch(err) {
+      logger.error('Could not eval "isReportCounted" function for (report=' + context.report._id + ', latestReport=' + context.latestReport._id +
+        '). Report will not be counted. Function passed: "' +
+        isReportCountedString + '". Error: ' +
+        err.message);
```
garethbowen: Use ES6 templating, eg:

```js
logger.error(`Could not eval "isReportCounted" function for (report=${context.report._id}, latestReport=...`);
```

transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,219 @@
...
+          return shouldBeAString;
+        });
+      }
+      return { error: 'multi_form_alerts : phone number for "' + recipient +
+        '" is not a string or array of strings. Message will not be sent. Found: "' +
+        JSON.stringify(evaled) + '"' };
+    } catch(err) {
+      return { error: 'multi_form_alerts : Could not find a phone number for "' + recipient +
+        '". Message will not be sent. Error: "' + err.message + '"' };
+    }
```
garethbowen: Use ES6 templating all through this function.

transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,219 @@
...
+    throw new Error('Bad config for multi_form_alerts. Expecting "numReportsThreshold" to be an integer. ' +
+      'E.g "numReportsThreshold": "3"');
+  }
+  if(!_.isArray(alert.recipients)) {
+    throw new Error('Bad config for multi_form_alerts. Expecting "recipients" to be an array of strings. ' +
+      'E.g "recipients": ["+9779841452277", "countedReports[0].contact.phone"]');
+  }
+  if (alert.forms && (!_.isArray(alert.forms))) {
+    alert.forms = null;
+    logger.warn('Bad config for multi_form_alerts. Expecting "forms" to be an array of form codes. Continuing without "forms", since it\'s optional.');
```
garethbowen: Lots more ES6 templating here!

transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,219 @@
...
+      if (countedReports.length >= alert.numReportsThreshold) {
+        return generateMessages(alert.recipients, alert.message, countedReports);
+      }
+      return false;
+    });
+};
+
+const onMatch = (change, db, audit, callback) => {
+  const latestReport = change.doc;
+  const alerts = getAlerts();
```
garethbowen: `config` or `alertsConfig` would be a better name IMO.

transitions/multi_form_alerts.js
```diff
@@ -0,0 +1,219 @@
...
+  const errors = [];
+  let docNeedsSaving = false;
+  let promiseSeries = Promise.resolve();
+  alerts.forEach((alert) => {
+    promiseSeries = promiseSeries.then(() => {
+      validateConfig(alert);
```
garethbowen: Didn't we decide to do this in some transition init rather than onMatch?

test/unit/transitions/multi_form_alerts.js
estellecomment: Alright, making a function to make it easier, and using it in individual tests.

transitions/multi_form_alerts.js
estellecomment: Good point. Filed #3631.

test/unit/transitions/multi_form_alerts.js
garethbowen: The new function is really unclear too. I'll change it and show you what I mean.

PR #139, opened 2017-07-07
Loops over transitions for signature tests

garethbowen:
# Description

Instead of each transition having unit tests to check the required
functions signatures match the expected, this change does all that
in a single loop.

medic/medic-webapp#3630

# Review checklist

### DIFF
--- a/transitions/index.js
+++ b/transitions/index.js
@@ -380,8 +380,6 @@ const attach = () => {
 };
 
 module.exports = {
-  _loadTransition: loadTransition,
-  _changeQueue: changeQueue,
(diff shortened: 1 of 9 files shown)


transitions/index.js
```diff
@@ -390,6 +388,10 @@ module.exports = {
   applyTransitions: applyTransitions
 };
 
-if (!process.env.TEST_ENV) {
+if (process.env.TEST_ENV) {
```
mandric: 👍

mandric:
I think it's already stated somewhere that transitions require these functions, so I don't think we need to document anything for this.

PR #140, opened 2017-07-07
Fixing npm ls by removing latest

SCdF:
# Description

Fixing npm ls by removing latest

medic/medic-webapp#3628

# Review checklist

- [x] Webapp CI runs clean against this branch


garethbowen:
Interestingly the package-lock has been massively updated indicating to me that `latest` does even get recorded in the lock file. This makes the error message much more sensible because essentially the lock is broken for those packages.

PR #141, opened 2017-07-07
Support both tests and prod medic/medic-webapp#3630

SCdF:
# Description

Recent changes broke prod because _loadTransition is not available

medic/medic-webapp#3630

# Review checklist

- [x] Webapp CI runs clean against this branch

### DIFF
--- a/transitions/index.js
+++ b/transitions/index.js
@@ -125,9 +125,7 @@ const loadTransitions = (autoEnableSystemTransitions = true) => {
       return logger.warn(`transition ${transition} is disabled`);
     }
 
-    // Needed to support tests mocking out _loadTransition, and _loadTransition
-    // not being available in prod.
-    (self._loadTransition || self.loadTransition)(transition);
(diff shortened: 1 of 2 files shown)


transitions/index.js
```diff
@@ -125,7 +125,9 @@ const loadTransitions = (autoEnableSystemTransitions = true) => {
       return logger.warn(`transition ${transition} is disabled`);
     }
 
-    self._loadTransition(transition);
+    // Needed to support tests mocking out _loadTransition, and _loadTransition
+    // not being available in prod.
+    (self._loadTransition || self.loadTransition)(transition);
```
mandric: Damn ugly, maybe we should go back to the old way? Where `_` means it's private use with caution.

mandric:
Gareth tried to introduce a new pattern, which I thought i liked, that only exports using underscore if we're running tests but we also need to reference the exported "private" function within the module during tests, so it can be mocked. Ugh

PR #142, opened 2017-07-13
Adds init function to validate transitions

garethbowen:
# Description

If this optional function is exported by the transition it's
called when the transition is loaded. Transition goading occurs
on start and on configuration change.

If the init function throws an error then the transition isn't
loaded, the error message is logged, and the other transitions
load as per usual.

medic/medic-webapp#3585

# Review checklist

### DIFF
--- a/transitions/index.js
+++ b/transitions/index.js
@@ -140,7 +140,11 @@ const loadTransitions = (autoEnableSystemTransitions = true) => {
 const loadTransition = key => {
   try {
     logger.info(`loading transition ${key}`);
-    transitions[key] = require('./' + key);
+    const transition = require('./' + key);
+    if (transition.init) {
+      transition.init();
+    }
+    transitions[key] = transition;
   } catch(e) {
     logger.error(`failed loading transition ${key}`);
     logger.error(e);
--- a/transitions/multi_form_alerts.js
+++ b/transitions/multi_form_alerts.js
@@ -6,7 +6,14 @@ const vm = require('vm'),
       messages = require('../lib/messages'),
       utils = require('../lib/utils'),
       transitionUtils = require('./utils'),
-      NAME = 'multi_form_alerts';
+      NAME = 'multi_form_alerts',
+      requiredFields = [
+        'isReportCounted',
+        'numReportsThreshold',
+        'message',
+        'recipients',
+        'timeWindowInDays'
+      ];
 
 const getAlertConfig = () => config.get('multi_form_alerts');
 
@@ -39,7 +46,7 @@ const countReports = (reports, isReportCountedString) => {
   return reports.filter((report) => {
     const context = { report: report, latestReport: reports[0]};
     try {
-      return vm.runInNewContext('(' + isReportCountedString + ')(report, latestReport)', context);
+      return vm.runInNewContext(`(${isReportCountedString})(report, latestReport)`, context);
     } catch(err) {
       logger.error(`Could not eval "isReportCounted" function for (report=${context.report._id}, latestReport=${context.latestReport._id}` +
         `). Report will not be counted. Function passed: "${isReportCountedString}". Error: ${err.message}`);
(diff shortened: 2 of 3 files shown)


SCdF:
> Transition goading occurs on start

You may wish to be careful here. Transitions look like peaceful creatures most of the time, but if provoked they can be quite unpredictable!

transitions/multi_form_alerts.js
```diff
@@ -39,7 +46,7 @@ const countReports = (reports, isReportCountedString) => {
   return reports.filter((report) => {
     const context = { report: report, latestReport: reports[0]};
     try {
-      return vm.runInNewContext('(' + isReportCountedString + ')(report, latestReport)', context);
+      return vm.runInNewContext(`(${isReportCountedString})(report, latestReport)`, context);
```
SCdF: 💯

transitions/multi_form_alerts.js
```diff
@@ -125,33 +132,33 @@ const getPhonesWithDuplicates = (recipients, countedReports) => {
...
-  if (alert.forms && (!_.isArray(alert.forms))) {
-    alert.forms = null;
-    logger.warn('Bad config for multi_form_alerts. Expecting "forms" to be an array of form codes. Continuing without "forms", since it\'s optional.');
-  }
+const validateConfig = () => {
+  const alertConfig = getAlertConfig();
+  alertConfig.forEach(alert => {
+    requiredFields.forEach(field => {
+      if (!alert[field]) {
+        throw new Error(`Bad config for multi_form_alerts. Expecting fields: ${requiredFields.join(', ')}`);
```
SCdF: This is really good. I think we can improve it by being more explicit in what is going wrong, to aid tech leads in fixing the problem. We could add the specific config that failed, for example. I think the best approach would be to refactor this to run through all configs and all fields, and then output the entire missing set.

From personal experience this is really helpful. It's always frustrating to fix one issue, run it through and then discover the next issue, rinse repeat etc.

transitions/multi_form_alerts.js
```diff
@@ -125,33 +132,33 @@ const getPhonesWithDuplicates = (recipients, countedReports) => {
   return getPhonesOneRecipient(recipients, countedReports);
 };
 
-const validateConfig = (alert) => {
```
SCdF: OT: what is up with this diff, github renders it really oddly? Was it tabs before and now it's spaces or something?

transitions/index.js
```diff
@@ -140,7 +140,11 @@ const loadTransitions = (autoEnableSystemTransitions = true) => {
 const loadTransition = key => {
   try {
     logger.info(`loading transition ${key}`);
-    transitions[key] = require('./' + key);
+    const transition = require('./' + key);
+    if (transition.init) {
```
SCdF: I know this is the existing behaviour, but now might be a good time to take a step back and re-consider our approach in allowing Sentinel to run with misconfigured transitions.

IMO we should not be doing this, because there is a good chance people are not going to be following the logs closely enough to realise that a transition of theirs is not running. I realise they will probably test this the first time they run it, but I imagine they will not re-test on subsequent updates.

It's unfortunate that we don't have a reliable way to exposing these sorts of problems in sentinel like we do in api (ie we can just refuse to boot if we have issues there). 

Interested in your thoughts here, and also perhaps @henokgetachew's as well. I wonder how much effort it would be to quickly whip up a status endpoint for sentinel, and expose that to whatever monitoring software we have.

I don't want to pile on extra work solving this one problem. I can imagine the monitoring could be a second ticket. I am interested in how you feel about having Sentinel fail to boot on bad config as a part of this PR though.

transitions/multi_form_alerts.js
```diff
@@ -208,5 +214,6 @@ module.exports = {
     doc.type === 'data_record' &&
     !transitionUtils.hasRun(doc, NAME)
   ),
-  onMatch: onMatch
+  onMatch: onMatch,
+  init: validateConfig
```
SCdF: I don't want to add more on to your plate necessarily, but do you think this is a good opportunity to do a once-over of existing transitions and see if there are quick `init`s we can add to those? Either here on via a raised ticket.

transitions/multi_form_alerts.js
garethbowen: It now validates all alerts, not just one alert, so the whole function body is now in a loop so it's 4 spaces instead of 2.

transitions/multi_form_alerts.js
garethbowen: I'll do an audit today and raise issues for any candidates I find.

transitions/multi_form_alerts.js
garethbowen: Fixed.

garethbowen:
Documentation added to https://github.com/medic/medic-docs/blob/master/development/transitions.md

PR #143, opened 2017-07-13
Multi form alerts transition performance

garethbowen:
# Description

Makes three performance improvements to the multi-form-alerts transition.

Firstly, docs are hydrated in one batch fetch from the database instead of
serially one at a time.

Secondly, the Script it parsed once per execution rather than once per report.
This could be further improved by parsing and caching once per configuration
change if necessary.

Thirdly, the report fetch is batched so we don't request too many docs in one
go.

medic/medic-webapp#3631

# Review checklist

### DIFF
--- a/lib/lineage.js
+++ b/lib/lineage.js
@@ -83,8 +83,75 @@ const fetchHydratedDoc = id => {
   });
 };
 
+const findRow = (rows, id) => id && rows.find(row => row.id === id);
+
+const collectContactIds = docs => {
+  const ids = [];
+  docs.forEach(doc => {
+    const contactId = doc.contact && doc.contact._id;
+    if (contactId && ids.indexOf(contactId) === -1) {
+      ids.push(contactId);
+    }
+    let parent = doc.parent;
+    while (parent) {
+      if (parent._id && ids.indexOf(parent._id) === -1) {
+        ids.push(parent._id);
+      }
+      parent = parent.parent;
+    }
+  });
+  return ids;
+};
+
+const hydrateDocs = docs => {
+  const ids = collectContactIds(docs);
+  if (!ids.length) {
+    return Promise.resolve();
+  }
+  return new Promise((resolve, reject) => {
+    db.medic.fetch({ keys: ids }, (err, results) => {
+      if (err) {
+        return reject(err);
+      }
+      const rows = results.rows;
+      docs.forEach(doc => {
+        const row = findRow(rows, doc.contact && doc.contact._id);
+        if (row && row.doc) {
+          doc.contact = row.doc;
+        }
+        let parent = doc;
+        while (parent) {
+          const row = findRow(rows, parent.parent && parent.parent._id);
+          if (row && row.doc) {
+            parent.parent = row.doc;
+          }
+          parent = parent.parent;
+        }
+      });
+      resolve();
+    });
+  });
+};
+
 module.exports = {
+  /**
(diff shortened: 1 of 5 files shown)


transitions/multi_form_alerts.js
```diff
@@ -153,22 +142,45 @@ const validateConfig = (alert) => {
   }
 };
 
+const count = (alert, latestReport) => {
+  return new Promise((resolve, reject) => {
```
SCdF: FWIW you can just do: `const count  = (alert, latestReport) => new Promise((resolve, reject) => {` here

transitions/multi_form_alerts.js
```diff
@@ -153,22 +142,45 @@ const validateConfig = (alert) => {
   }
 };
 
+const count = (alert, latestReport) => {
```
SCdF: I think it would be helpful to either write a short comment explaining what this function does, or rename it to something less generic than `count`

lib/lineage.js
```diff
@@ -83,8 +83,75 @@ const fetchHydratedDoc = id => {
   });
 };
 
+const findRow = (rows, id) => id && rows.find(row => row.id === id);
```
garethbowen: It is. I'm avoiding underscore where possible.

lib/lineage.js
```diff
@@ -83,8 +83,75 @@ const fetchHydratedDoc = id => {
...
+
+const collectContactIds = docs => {
+  const ids = [];
+  docs.forEach(doc => {
+    const contactId = doc.contact && doc.contact._id;
+    if (contactId && ids.indexOf(contactId) === -1) {
```
garethbowen: I need to find a better resource to find out what's supported in Node versions. Any suggestions?

lib/lineage.js
```diff
@@ -83,8 +83,75 @@ const fetchHydratedDoc = id => {
...
+  return ids;
+};
+
+const hydrateDocs = docs => {
+  const ids = collectContactIds(docs);
+  if (!ids.length) {
+    return Promise.resolve();
+  }
+  return new Promise((resolve, reject) => {
+    db.medic.fetch({ keys: ids }, (err, results) => {
```
garethbowen: Sounds great. But not today...

lib/lineage.js
```diff
@@ -83,8 +83,75 @@ const fetchHydratedDoc = id => {
...
+   * Given a doc id get a doc and all parents and contacts
+   * @param id The id of the doc
+   * @returns Promise
+   */
   fetchHydratedDoc: fetchHydratedDoc,
+
+  /**
+   * Given an array of minified docs bind the parents and contacts
+   * @param docs The array of docs to hydrate
+   * @returns Promise
```
garethbowen: http://usejsdoc.org/

transitions/multi_form_alerts.js
garethbowen: Yeah. I don't think that's more readable though...

garethbowen: Any time I have more than one `=>` on a line I get confused  :(

lib/lineage.js
SCdF: Why? Is this a code-wide decision that I missed?

lib/lineage.js
SCdF: It's not perfect, but http://node.green is what I use. From that you can see it's been in there since 6.4:
image

PR #146, opened 2017-07-21
Stops processing changes if config is invalid

garethbowen:
# Description

Transitions now run validation via the `init` function on start up
and config change. If any transition init throws an Error then no
further changes will be processed but processing will resume once
the configuration is fixed.

medic/medic-webapp#3668

# Review checklist

### DIFF
--- a/config.js
+++ b/config.js
@@ -29,13 +29,7 @@ const initFeed = () => {
   feed.on('change', change => {
     if (change.id === '_design/medic') {
       logger.info('Reloading configuration');
-      initConfig(err => {
-        if (err) {
-          console.error('Error loading configuration. Exiting...');
-          process.exit(0);
-        }
-        require('./transitions').loadTransitions();
-      });
+      initConfig();
     } else if (change.id.startsWith('messages-')) {
       logger.info('Detected translations change - reloading');
       loadTranslations();
@@ -44,10 +38,11 @@ const initFeed = () => {
   feed.follow();
 };
 
-const initConfig = callback => {
+const initConfig = () => {
   db.medic.get(SETTINGS_PATH, (err, data) => {
     if (err) {
-      return callback(err);
+      console.error('Error loading configuration. Exiting...');
+      process.exit(0);
     }
     _.defaults(data.settings, defaults);
     config = data.settings;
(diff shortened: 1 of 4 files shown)


config.js
```diff
@@ -58,7 +53,7 @@ const initConfig = callback => {
       config.schedule_evening_hours,
       config.schedule_evening_minutes
     );
-    callback();
+    require('./transitions').loadTransitions();
```
SCdF: Is there a reason we require this every time btw? AFAICT require doesn't do anything after the first go:
```js
> require('./requireme')
You loaded me!
{}
> require('./requireme')
{}
```

So it really just seems to be a shortcut if you don't want to create a variable to store the reference in.

transitions/index.js
```diff
@@ -15,6 +15,8 @@ const _ = require('underscore'),
       PROGRESS_REPORT_INTERVAL = 500, // items
       transitions = {};
 
+let feed;
```
SCdF: `feed` doesn't really mean anything to me. However, this is probably because I'm not familiar with the follow.feed function. Maybe `changesFeed` would make more sense?

Anyway, it's not hard to work out what it does by just looking for `feed =` in the codebase. I'll leave it to you if you want to consider a rename.

transitions/index.js
```diff
@@ -135,20 +144,22 @@ const loadTransitions = (autoEnableSystemTransitions = true) => {
       return logger.warn(`transition ${key} not available.`);
     }
   });
+
+  if (loadError) {
+    logger.error(`Transitions are disabled until the above configuration errors are fixed.`);
```
SCdF: Because this comes *after*:
```js
return logger.warn(`transition ${key} not available.`);
```
I'd be worried that people may think that these warns are part of what they need to fix.

config.js
garethbowen: Because otherwise all the tests break. I spent some time trying to work out why this would break all the tests but it's not worth me spending any more time on it now.

transitions/index.js
garethbowen: Done.

transitions/index.js
garethbowen: They really should fix the warn parts - it means they have configuration that they think is doing something and it's not. At best it's a failed attempt at loading an unreleased transition, at worst it's a typo in a transition name and they're not running a transition they think they are.

I've upgraded it to an error so sentinel transitions won't load if an unknown name is configured.

PR #148, opened 2017-07-26
Batching multi_form_alerts' getPhones and generateMessages

estellecomment:
# Description

Convert getPhones processing to use batches of reports in multi_form_alerts, to avoid overly big sets of records.
Limit the available information for generateMessages, to avoid memory problems. Only message about newReports that haven't been messaged about before.

medic/medic-webapp#3678

# Review checklist

### DIFF
--- a/transitions/multi_form_alerts.js
+++ b/transitions/multi_form_alerts.js
@@ -44,9 +44,8 @@ const countReports = (reports, latestReport, script) => {
   });
 };
 
-const generateMessages = (alert, countedReports) => {
+const generateMessages = (alert, phones, countedReports) => {
   let isLatestReportChanged = false;
-  const phones = getPhones(alert.recipients, countedReports);
   phones.forEach((phone) => {
     if (phone.error) {
       logger.error(phone.error);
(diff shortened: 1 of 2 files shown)


transitions/multi_form_alerts.js
```diff
@@ -169,42 +182,73 @@ const validateConfig = () => {
...
+ * Returns { countedReportsIds, newReports, phones }.
+ */
+const getCountedReportsAndPhones = (alert, latestReport) => {
   return new Promise((resolve, reject) => {
     const script = vm.createScript(`(${alert.isReportCounted})(report, latestReport)`);
-    let reports = countReports([ latestReport ], latestReport, script);
     let skip = 0;
+    let countedReportsIds = [ latestReport._id ];
+    let newReports = [ latestReport ];
+    let phones = [ ...getPhonesOneReport(alert.recipients, latestReport) ];
```
garethbowen: Wouldn't this be equivalent:

```
let phones = getPhonesOneReport(alert.recipients, latestReport);
```

transitions/multi_form_alerts.js
```diff
@@ -169,42 +182,73 @@ const validateConfig = () => {
...
   });
 };
 
+/**
+ * Returns Promise({ numFetched, countedReportsIds, newReports, phones }) for the db batch with skip value.
+ */
+const getCountedReportsAndPhonesBatch = (script, latestReport, alert, skip) => {
+  const options = { skip: skip, limit: BATCH_SIZE };
+  const output = {};
+  let countedReports;
```
garethbowen: This can be declared when you instantiate it thus reducing its scope I think.

transitions/multi_form_alerts.js
```diff
@@ -169,42 +182,73 @@ const validateConfig = () => {
...
+      return countedReports;
+    })
+    .then(counted => {
+      output.newReports = counted.filter(report => !isReportAlreadyMessaged(report, alert.name));
+      return output.newReports;
+    })
+    .then((newReports) => {
+      output.phones = getPhones(alert.recipients, newReports);
+      return output;
+    });
```
garethbowen: Sometimes I like breaking sync code up into different `.then` blocks but I think in this case it makes it more confusing because you have the `output` object being updated in each. It'd be clearer to do...

```js
  return fetchReports(latestReport.reported_date - 1, alert.timeWindowInDays, alert.forms, options)
    .then(fetched => {
      const countedReports = countReports(fetched, latestReport, script);
      const newReports = countedReports.filter(report => !isReportAlreadyMessaged(report, alert.name));
      return {
        numFetched: fetched.length,
        countedReportsIds: countedReports.map(report => report._id),
        newReports: newReports,
        phones: getPhones(alert.recipients, newReports)
      };
    });
```

transitions/multi_form_alerts.js
estellecomment: how about `[...[...getPhonesOneReport()]]`, even cooler

transitions/multi_form_alerts.js
estellecomment: 👍

estellecomment:
Thanks for the cleanup! Merging.

PR #155, opened 2017-08-08
Update to use contacts_by_reference view

garethbowen:
# Description

Use a new combined view to get contacts by shortcode or refid.

medic/medic-webapp#3511

# Review checklist

### DIFF
--- a/lib/ids.js
+++ b/lib/ids.js
@@ -99,8 +99,8 @@ const putIdLengthDoc = (db, idLengthDoc) =>
  */
 const findUnusedIds = (db, freshIds) =>
   new Promise((resolve, reject) => {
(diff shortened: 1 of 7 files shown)


garethbowen:
This change is part of merging three views into one. The simprints part has to be available client side. My assertion is that the overhead of emitting refids and shortcodes is negligible because the real cost is in fetching the doc and compiling the view. This change saves us on server side (in particular, view rebuilding on release) by having 1 view instead of 3.

alxndrsn:
> the overhead of emitting refids and shortcodes is negligible

I think this is a valid opinion, but worth noting in comments on the view so that it's clear that although parts of that view are _included_ in `medic-client`, they're not currently for use by the "client".

PR #156, opened 2017-08-08
Scheduled messages support person in message template

SCdF:
# Description

Extend support for the person document to scheduled messages as well as standard messages.
Support the same legacy hook for `{{patient_name}}`.

medic/medic-webapp#3419

# Review checklist

### DIFF
--- a/lib/messages.js
+++ b/lib/messages.js
@@ -3,6 +3,34 @@ var _ = require('underscore'),
     utils = require('./utils'),
     logger = require('./logger');
 
+function extendedTemplateContext(doc, extras) {
+    const templateContext = module.exports.extractTemplateContext(doc);
+
+    if (extras.templateContext) {
+        _.defaults(templateContext, extras.templateContext);
+    }
+
+    if (extras.patient) {
+        _.defaults(templateContext, module.exports.extractTemplateContext(extras.patient));
+    }
+
+    if (extras.registrations && extras.registrations.length) {
+        _.defaults(templateContext, module.exports.extractTemplateContext(extras.registrations[0].doc));
+    }
+
+    if (extras.person) {
+        templateContext.person = extras.person;
+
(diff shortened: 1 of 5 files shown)


lib/messages.js
```diff
@@ -3,6 +3,34 @@ var _ = require('underscore'),
     utils = require('./utils'),
     logger = require('./logger');
 
+function extendedTemplateContext(doc, extras) {
```
estellecomment: arrow?

estellecomment: nit : can you either document what fields are expected in `extras`, or pass them as individual arguments? I'm fighting against the extras/options pattern, it makes things hard to follow. (types!)

transitions/registration.js
```diff
@@ -359,23 +359,25 @@ module.exports = {
...
+
+    async.parallel({
+      registrations: _.partial(getRegistrations, options.db, patientId),
+      person: _.partial(utils.getPatientContact, options.db, patientId)
+    }, (err, {registrations, person}) => {
+      if (err) {
+        return callback(err);
       }
-    );
+      options.params.forEach(name => {
```
estellecomment: not your code, but `name` is a pretty bad name, I have no idea what this is

lib/messages.js
SCdF: You mean arrow function? I'm following the convention in the rest of the file. Some day it will be converted to ES6, but not today!

SCdF: I considered moving the documentation from the function below, but I figured since the function is basically just a bunch of `if (extras.BLAH)` statements it was obvious enough?

transitions/registration.js
SCdF: I'm not sure just reading it what it does either! I'll look into the code and see if I can name it better. The whole interaction with transition config can be really confusing..

PR #157, opened 2017-08-16
Use new_report in multi_form_alert recipient context.

estellecomment:
# Description
Rename variable, including in the context used to evaluate the recipients. 
medic/medic-webapp#3715

# Review checklist

### DIFF
--- a/transitions/multi_report_alerts.js
+++ b/transitions/multi_report_alerts.js
@@ -85,8 +85,8 @@ const generateMessages = (alert, phones, latestReport, countedReportsIds, newRep
 // Recipients format examples:
 // [
 //    '+254777888999',
-//    'countedReport.contact.parent.parent.contact.phone',   // returns string
-//    'countedReport.contact.parent.parent.alertRecipients', // returns string array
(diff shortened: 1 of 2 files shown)


transitions/multi_report_alerts.js
```diff
@@ -85,8 +85,8 @@ const generateMessages = (alert, phones, latestReport, countedReportsIds, newRep
 // Recipients format examples:
 // [
 //    '+254777888999',
-//    'countedReport.contact.parent.parent.contact.phone',   // returns string
-//    'countedReport.contact.parent.parent.alertRecipients', // returns string array
+//    'new_report.contact.parent.parent.contact.phone',   // returns string
```
abbyad: Just a comment, but this should have the item, right? eg `new_report.0.contact`...

estellecomment: nope, because you specify recipients for each `new_report`, you don't have access to the whole array (it's batched for performance, so you have a chunk of the `new_reports` array, not the whole thing)

PR #159, opened 2017-08-17
Use the correct function name

SCdF:
Nano, not PouchDB!

medic/medic-webapp#3790
medic/medic-projects#2648

# Review checklist

### DIFF
--- a/lib/ids.js
+++ b/lib/ids.js
@@ -79,7 +79,7 @@ const getIdLengthDoc = db =>
 
 const putIdLengthDoc = (db, idLengthDoc) =>
   new Promise((resolve, reject) =>
-    db.medic.put(idLengthDoc, err => {
+    db.medic.insert(idLengthDoc, err => {
       if (err) {
         // We're OK with a 409, because we're going to presume this is happening
         // because a human edited it to suite their needs, and their write is more
--- a/test/unit/ids.js
+++ b/test/unit/ids.js
@@ -13,7 +13,7 @@ const mockDb = (idFilterLogicFn) => {
       callback(null, toReturn);
     }),
     get: sinon.stub().callsArgWith(1, null, {_id: 'shortcode-id-length', current_length: 5}),
-    put: sinon.stub().callsArgWith(1)
+    insert: sinon.stub().callsArgWith(1)
   }};
 };
 


SCdF:
A note on testing. The way this would have been found would have been through more extensive manual testing (which is my fault for not doing) by generating enough documents to force it to bump the number.

It would also have been found with integration testing. Unfortunately we don't have a framework for this in sentinel right now.

It is a valid question as to whether we or do not want to introduce this framework.

SCdF:
Actually @estellecomment don't review this yet, I've found more problems due to recent changes

SCdF:
I've tested and confirmed that this works.

I wrote a small custom script you run from the root of sentinel, on an empty but initialised (with ddoc) DB, after modifying the `MIN_ID_LENGTH` in ids.js to `3` (So it only has to generate ~1000 docs as opposed to ~100000):

```js
const db = require('./db'),
      ids = require('./lib/ids');

const generator = ids.generator(db);

const doc = id => {return {
  type: 'person',
  patient_id: id,
  reported_date: new Date()
};};

const ITERATIONS = 10000;
let   iteration = 0;

const promiseLoop = () => {
  if (ITERATIONS !== ++iteration) {
    return loop().then(promiseLoop);
  }
};

const loop = () => generator.next().value.then(patientId => {
  const nextDoc = doc(patientId);
  console.log(`About to insert ${nextDoc.patient_id}`);
  return new Promise((resolve, reject) => db.medic.insert(nextDoc, (err, result) => {
    if (err) {
      reject(err);
    } else {
      resolve(result);
    }
  }));
});

return promiseLoop()
  .catch(console.error);
```

If you run this you can see it doing stuff like this:

```
About to insert 700
About to insert 036
About to insert 760
About to insert 784
About to insert 267
warn: Could not create a unique id of length 3, increasing length
About to insert 5903
About to insert 8310
About to insert 3773
About to insert 8941
About to insert 2343
```

I'm not sure how to formalise this kind of thing honestly. We don't want to run this test every time, even if I could re-shape it into an integration test.

estellecomment:
Why don't you want to run that test every time? Because it's too long to run? `MIN_ID_LENGTH` = 1 would make it short enough, just loop it until it collides and check the next id has length > 1.

But yeah we don't have integration tests setup for sentinel yet. Would be nice, for transitions too, this kind of wrong db mocking is pretty common...

PR #160, opened 2017-08-17
Fix lineage.hydrateDocs to use the right fields for data_records. 

estellecomment:
# Description
Edit the bulk hydration of docs to using the right doc structure for data_records. They don't have a parent field.
See test's `setUp` for the right data structure. Refactored the test to reuse that data (we had the single-doc function doing something different than the bulk function...)

medic/medic-webapp#3789

# Review checklist

### DIFF
--- a/lib/lineage.js
+++ b/lib/lineage.js
@@ -1,20 +1,25 @@
-const db = require('../db');
+const _ = require('underscore'),
+    db = require('../db');
 
-const bindContacts = (lineage, contacts) => {
+const fillContactsInDocs = (docs, contacts) => {
+  if (!contacts || !contacts.length) {
+    return docs;
+  }
   contacts.forEach(contactDoc => {
-    const stub = lineage.find(lineageDoc => {
-      const id = lineageDoc &&
-                 lineageDoc.contact &&
-                 lineageDoc.contact._id;
+    const stub = docs.find(doc => {
+      const id = doc &&
+                 doc.contact &&
+                 doc.contact._id;
       return id === contactDoc._id;
     });
     if (stub) {
       stub.contact = contactDoc;
     }
   });
+  return docs;
 };
 
-const bindParents = (lineage, doc) => {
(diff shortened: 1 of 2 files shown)


lib/lineage.js
```diff
@@ -68,75 +73,133 @@ const fetchHydratedDoc = id => {
...
-        while (parent) {
-          const row = findRow(rows, parent.parent && parent.parent._id);
-          if (row && row.doc) {
-            parent.parent = row.doc;
-          }
-          parent = parent.parent;
+    while (current) {
+      if (current.parent && current.parent._id) {
+        /*jshint loopfunc:true */
+        const parentDoc = parents.find(parent => parent._id === current.parent._id);
```
garethbowen: To placate jshint, pull out a function, eg:

```js
const findById = (id, docs) => docs.find(doc => doc._id === id);
```

Or if that feels stupid, add a comment on what "loopfunc" is and why it's not an issue here (and I agree it's not).

test/unit/lineage.js
```diff
@@ -2,510 +2,514 @@ const db = require('../../db'),
...
+  test.equals(hydratedReport.contact.parent.parent.contact.name, report_grandparentContact.name);
+
+  test.equals(hydratedPlace.contact.name, place_contact.name);
+  test.equals(hydratedPlace.parent.name, place_parent.name);
+  test.equals(hydratedPlace.parent.contact.name, place_parentContact.name);
+  test.equals(hydratedPlace.parent.parent.name, place_grandparent.name);
+  test.equals(hydratedPlace.parent.parent.contact.name, place_grandparentContact.name);
+
+  test.done();
+}).catch(test.done);
```
estellecomment: no, it fails when you pass an arg to `test.done`. Saw it in action.

garethbowen: Ugh... I must be thinking of karma. Sorry...

PR #163, opened 2017-08-24
Schedules are messing with each other due to bad management of groups

estellecomment:
# Description

Silence schedules one by one rather than all together, to fix a bug where schedules' clearing was influencing other schedules' clearing. 
See test that fails because of bug : https://github.com/medic/medic-sentinel/pull/163/commits/bca9411c64e3c87a3284b5f314079a11b8c3dae6

medic/medic-webapp#3824

# Review checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -304,6 +304,61 @@ const getPatient = (db, patientShortcodeId, includeDocs, callback) => {
   });
 };
 
+/*
+ * type can be array or string
+ */
+const getScheduledTasksByType = (doc, type) => {
+  const scheduled_tasks = doc && doc.scheduled_tasks;
+  return _.filter(scheduled_tasks, task => {
+    if (_.isArray(type)) {
+      return type.indexOf(task.type) >= 0;
+    }
+    return task.type === type;
+  });
+};
+
+/*
+ * Return false when the recipient phone matches the denied list.
+ *
+ * outgoing_deny_list is a comma separated list of strings. If a string in
+ * that list matches the beginning of the phone then we set up a response
+ * with a denied state. The pending message process will ignore these
+ * messages and those reports will be left without an auto-reply. The
+ * denied messages still show up in the messages export.
+ *
+ * @param {String} from - Recipient phone number
+ * @returns {Boolean}
+ */
+const isOutgoingAllowed = from => {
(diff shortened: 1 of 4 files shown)


estellecomment:
There's a lot of refactoring, so reviewing will be tricky. 
 - rename vars and funcs
 - remove the `options` argument to functions, because it's hard to tell what's in it, so the code is hard to read and reuse, and some things are passed which actually aren't needed.
 - de-expose functions : only expose functions in `modules.exports` if they're used outside of the module. This caused changes to the tests, because some exposed functions were exposed to be mocked for testing, in order to test sub-parts of the module. I changed most of these tests to test the whole flow instead. (`_silenceReminders` is left exposed, too much changes needed)
 - split the code into functions that are sub-routines, instead of functions that are chopping a large function into first-part-second-part-third-part.
 - ES6 : remove some use of underscore, and add `const` and `=>`
 - create vars where needed, rather than at top of function.

I suggest reviewing in 3 steps : 
 - refactoring commits : https://github.com/medic/medic-sentinel/pull/163/files/57d124610e7f0b91ed4cd56577a26c3cd2fb9e7d
 - adding of the failing test (second-to-last commit): https://github.com/medic/medic-sentinel/pull/163/commits/bca9411c64e3c87a3284b5f314079a11b8c3dae6
 - fixing of the code, making the test pass (last commit) : https://github.com/medic/medic-sentinel/pull/163/commits/34f1467816fc3943fa1e09e9b7b35011d18672d2

lib/utils.js
```diff
@@ -304,6 +304,56 @@ const getPatient = (db, patientShortcodeId, includeDocs, callback) => {
   });
 };
 
+/*
+ * type is a string
+ */
+const getScheduledTasksByType = (doc, type) => {
+  const scheduled_tasks = doc && doc.scheduled_tasks;
+  return scheduled_tasks.filter(task => task.type === type );
```
mandric: This will throw if scheduled_tasks is undefined or null.

mandric:
Might be good to do some acceptance testing on this one since it's a decent sized refactor and kind of a delicate part of the code.

estellecomment:
Force-pushed with squashed commits for clean merge. (so links above are broken)

PR #164, opened 2017-08-25
Don't round birthdays or expected birthdays

alxndrsn:


transitions/registration.js
```diff
@@ -119,7 +119,7 @@ module.exports = {
     const today = moment(date.getDate()).startOf('day');
     const weeks = parseInt(module.exports.getWeeksSinceDOB(doc), 10);
     if (!isNaN(weeks)) {
-      return today.startOf('week').subtract(weeks, 'weeks');
+      return today.startOf('day').subtract(weeks, 'weeks');
```
garethbowen: `today` has already been set to `startOf('date');` on line 119 - no need to call it again.

PR #165, opened 2017-08-28
Don't escape template parameters

garethbowen:
# Description

Mustache escape HTML characters by default because usually it's
used for displaying HTML content so it's at risk for injection.
We just use it for generating SMS messages to HTML escaping isn't
relevant for us. We use angular to escaping the message content
when displayed in the webapp.

medic/medic-webapp#3795

# Review checklist

### DIFF
--- a/test/unit/messages.js
+++ b/test/unit/messages.js
@@ -144,6 +144,25 @@ exports['addMessage supports template variables on doc'] = function(test) {
     test.done();
 };
 
+exports['addMessage does not escape characters - #3795'] = function(test) {
+    var doc = {
+        form: 'x',
+        reported_date: '2050-03-13T13:06:22.002Z',
+        place: 'Sharon\'s Place'
+    };
+    messages.addMessage({
+        doc: doc,
+        phone: '+13125551212',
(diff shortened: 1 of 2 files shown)


test/unit/messages.js
```diff
@@ -144,6 +144,25 @@ exports['addMessage supports template variables on doc'] = function(test) {
     test.done();
 };
 
+exports['addMessage does not escape characters - #3795'] = function(test) {
+    var doc = {
+        form: 'x',
+        reported_date: '2050-03-13T13:06:22.002Z',
+        place: 'Sharon\'s Place'
```
alxndrsn: I'm surprised `'` actually gets escaped at all, but checking mustache source it does:
```
  var entityMap = {
    '&': '&amp;',
    '<': '&lt;',
    '>': '&gt;',
    '"': '&quot;',
    "'": '&#39;',
    '/': '&#x2F;',
    '`': '&#x60;',
    '=': '&#x3D;'
  };
```
Might be worth adding a couple of `>`, `<` or `&` to the example to make it more obvious.

PR #171, opened 2017-09-05
3857 fix refactor

SCdF:
# Description

Two main changes:
 - Fixed a function caller who was missed when the function signature was refactored
 - Changing the findToClear function correctly. See [this comment](https://github.com/medic/medic-webapp/issues/3824#issuecomment-327068751)

Specifically:
 - If there is no silence_for configured, mark all clearable tasks as cleared
 - If there is a silence_for configured
   - find the maximum visible date by extending the reports `reported_date` by `silence_for`
   - find all groups that "exist" behind that date, where exist means have at least one task with a due date before the date
   - and clear all clearable tasks for all those groups
 - We define clearable as any scheduled or pending task. Note that previously it was just scheduled, but adding pending makes sense, because a pending task is not any different from a scheduled one really, it's just that gateway and friends haven't picked it up yet (maybe gateway is down and has been down for weeks).

medic/medic-webapp#3857

# Review checklist


transitions/accept_patient_reports.js
```diff
@@ -13,52 +13,53 @@ const _hasConfig = (doc) => {
     return Boolean(getConfig(doc.form));
 };
 
+// This is more complicated than it needs to be because JS / _ always use ===
+// for equality, complicating the use of unique tuples (i.e. [1,2] !== [1,2])
+const hasGroupTypeCombo = (groupTypeColl, [group, type]) =>
```
garethbowen: I think `hasGroupAndType` might be a better name?

transitions/accept_patient_reports.js
```diff
@@ -13,52 +13,53 @@ const _hasConfig = (doc) => {
...
+    const tasksUnderReview = utils.getScheduledTasksByType(registration, taskTypes);
+
+    if (!config.silence_for) {
+        // No range, all clearable tasks should be cleared
+        return tasksUnderReview.filter(task => typesToClear.includes(task.state));
+    } else {
+        // Clear all tasks that are members of a group that "exists" before the
+        // silence_until date. e.g., they have at least one task in their group
+        // whose due date is before silence_until.
+        const silence_until = reportedDateMoment.clone();
```
garethbowen: Go camel case for variables.

SCdF: Heh, I didn't even notice. I wonder if it was originally called that because it used to be a config var? Anyway, changed.

transitions/accept_patient_reports.js
SCdF: Sure.

PR #173, opened 2017-09-13
Lineage bug, notes and test case

SCdF:


SCdF:
@garethbowen / @estellecomment interested in your opinions here.

Firstly: I want to make sure this is a valid test case. I'm pretty sure it is. I'm not sure why it clearly often works, but I presume it's related to ordering.

Second: I'm not sure how to fix this. If I start screwing with find and just mapping over things we get into circular reference territory, and I haven't rolled up my sleeves yet to understand exactly how all this clicks together and why this is happening. I think as well that perhaps the solution may be different if we fix the two optimisation FIXME's that are noted there. For example, if we leave the optimisations out, while we're pulling the same doc down twice, we *do* know that it should bind exactly twice, which could be helpful in not getting circular references. On the other hand, if we don't have to do that we should really fix it, because it's a waste of bandwidth / etc (and in some cases we might not even need to go to the database for contacts at all!).

SCdF:
Here is the failure:
```
>> Actual:
>>   { type: 'data_record',
>>     _id: 'docId',
>>     contact: 
>>      { type: 'person',
>>        _id: 'chwId',
>>        parent: 
>>         { type: 'clinic',
>>           _id: 'areaId',
>>           contact: { _id: 'chwId' },
>>           parent: undefined },
>>        hydrated: true } }
>> Operator:
>>   deepEqual
>> Expected:
>>   { _id: 'docId'9m,
>>     contact: 
>>      { _id: 'chwId',
>>        parent: { _id: 'areaId', contact: { _id: 'chwId', hydrated: true } },
>>        hydrated: true } }
```

Note that in the actual the `contact.parent.contact` remains undefined, whereas my expectation is that it would be hydrated.

PR #176, opened 2017-10-12
Delete server side info and meta docs

garethbowen:
# Description

When a deletion is detected remove the associated meta doc and all
associated read docs in admin meta dbs. The non-admin read docs
will be deleted client side to avoid sentinel having to delete docs
from potentially thousands of databases.

medic/medic-webapp#3944

# Review checklist

### DIFF
--- a/transitions/index.js
+++ b/transitions/index.js
@@ -79,13 +79,27 @@ const processChange = (change, callback) => {
   if (!change) {
     return callback();
   }
+  logger.debug(`change event on doc ${change.id} seq ${change.seq}`);
+  if (processed > 0 && (processed % PROGRESS_REPORT_INTERVAL) === 0) {
+    logger.info(`transitions: ${processed} items processed (since sentinel started)`);
+  }
+  if (change.deleted) {
+    // don't run transitions on deleted docs, but do clean up
+    async.parallel([
+      async.apply(deleteInfoDoc, change),
+      async.apply(deleteReadDocs, change)
+    ], err => {
+      if (err) {
+        logger.error('Error cleaning up deleted doc', err);
+      }
+      processed++;
+      updateMetaData(change.seq, callback);
+    });
+    return;
+  }
   lineage.fetchHydratedDoc(change.id)
     .then(doc => {
-      if (processed > 0 && (processed % PROGRESS_REPORT_INTERVAL) === 0) {
-        logger.info(`transitions: ${processed} items processed (since sentinel started)`);
-      }
       change.doc = doc;
-      logger.debug(`change event on doc ${change.id} seq ${change.seq}`);
       const audit = require('couchdb-audit')
             .withNano(db, db.settings.db, db.settings.auditDb, db.settings.ddoc, db.settings.username);
       module.exports.applyTransitions({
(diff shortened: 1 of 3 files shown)


transitions/index.js
```diff
@@ -103,6 +117,83 @@ const processChange = (change, callback) => {
...
+};
+
+const deleteReadDocs = (change, callback) => {
+
+  // we don't know if the deleted doc was a report or a message so
+  // attempt to delete both
+  const possibleReadDocIds = [ 'report', 'message' ]
+    .map(type => `read:${type}:${change.id}`);
+
+  getAdminNames((err, admins) => {
```
SCdF: Do we have access to the entire document at this stage? Ideally we'd only do this if the document type was `data_record`.

transitions/index.js
```diff
@@ -103,6 +117,83 @@ const processChange = (change, callback) => {
...
+      return callback(err);
+    }
+    doc._deleted = true;
+    db.medic.insert(doc, callback);
+  });
+};
+
+const getAdminNames = callback => {
+  const options = { key: [ 'user-settings' ], include_docs: true };
+  db.medic.view('medic-client', 'doc_by_type', options, (err, results) => {
```
SCdF: I am really wondering if a new view would be better here (`admin_users`). They view will be very small, and it means that we don't have to load over every user document every time a delete happens.

transitions/index.js
```diff
@@ -103,6 +117,83 @@ const processChange = (change, callback) => {
...
+    if (err) {
+      return callback(err);
+    }
+    const admins = results.rows
+      .filter(row => {
+        return row.doc.roles &&
+               (row.doc.roles.includes('_admin') ||
+               row.doc.roles.includes('national_admin'));
```
SCdF: I realise why we need to do this this way, but we really need a better way of determining whether or not someone is online-first or offline-first.

Can we add a TODO here and link that to a ticket to add this as a flag in the user document? Linking the TODO so we'd know to change this code later. Can discuss when you're online

transitions/index.js
```diff
@@ -103,6 +117,83 @@ const processChange = (change, callback) => {
...
+
+  getAdminNames((err, admins) => {
+    if (err) {
+      return callback(err);
+    }
+
+    async.each(
+      admins,
+      (admin, callback) => {
+        const metaDb = db.use(`medic-user-${admin}-meta`);
```
SCdF: Just to double-check (because I have no trust) this is stateless right? We're not modifying `db` with this call?

transitions/index.js
```diff
@@ -103,6 +117,83 @@ const processChange = (change, callback) => {
...
+              // no meta data db exists for this user - ignore
+              return callback();
+            }
+            return callback(err);
+          }
+          metaDb.fetch({ keys: possibleReadDocIds }, (err, results) => {
+            if (err) {
+              return callback(err);
+            }
+            const row = results.rows.find(row => !!row.doc);
```
SCdF: Not sure why you'd need to do this line. When would you have a row that didn't also contain a doc?

Oh, is this because it returns 404 rows for each key you fetch?

If so, maybe a comment would make this clearer to the reader?

transitions/index.js
garethbowen: I think so. The documentation says:

> creates a scope where you operate inside name.

https://github.com/dscape/nano#nanousename

This implies to me that the returned value is a scoped db.

transitions/index.js
garethbowen: Done.

transitions/index.js
garethbowen: Unfortunately, no. Just the id. And we can't fetch it either because it's just been deleted.

transitions/index.js
garethbowen: You're right - this could be a big deal for big databases. I'll fix it.

transitions/index.js
garethbowen: Created issue: https://github.com/medic/medic-webapp/issues/3960

I didn't add the TODO because the code here got moved into a view as per your previous comment.

PR #178, opened 2017-10-26
Fix sentinel fti integration

garethbowen:
# Description

The `db.fti` function wasn't working due to a double slash in the
URL. It's not clear when this was broken.

Added logging so when validation functions error out the error
is logged.

Also future-proofed the `db.fti` function using the similar code
from medic-api so it works with couchdb2.

medic/medic-webapp#651

# Review checklist

### DIFF
--- a/db.js
+++ b/db.js
@@ -1,14 +1,17 @@
 var nano = require('nano'),
 	url = require('url'),
-    path = require('path');
+    path = require('path'),
+    request = require('request');
 
 var couchUrl = process.env.COUCH_URL;
 if (couchUrl) {
     // strip trailing slash from to prevent bugs in path matching
     couchUrl = couchUrl.replace(/\/$/, '');
     var parsedUrl = url.parse(couchUrl);
+    var baseUrl = couchUrl.substring(0, couchUrl.indexOf('/', 10));
+    var luceneUrl = baseUrl.replace('5984', '5985');
 
-    module.exports = nano(couchUrl.substring(0, couchUrl.indexOf('/', 10)));
+    module.exports = nano(baseUrl);
     module.exports.medic = nano(couchUrl);
 
     var dbName = parsedUrl.path.replace('/','');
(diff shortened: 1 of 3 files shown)


db.js
```diff
@@ -1,14 +1,17 @@
 var nano = require('nano'),
 	url = require('url'),
-    path = require('path');
+    path = require('path'),
+    request = require('request');
```
SCdF: How has this ever worked without request before?

db.js
```diff
@@ -28,8 +31,38 @@ if (couchUrl) {
     }
 
     module.exports.fti = function(index, data, cb) {
-        var uri = path.join('/_fti/local', module.exports.settings.db, '_design', module.exports.settings.ddoc, index);
-        module.exports.request({ path: uri, qs: data }, cb);
```
SCdF: Did `module.exports.request` ever have a value? I'm so confused by this old code!

garethbowen: This line is confusing and easy to miss: `module.exports = nano(baseUrl);`. This means that `modules.exports.request` is set to the `request` function in nano. We then extend nano with our own properties like `fti` and `settings`. It's a terrible pattern IMO because it's really hard to read and you risk overwriting something important in nano.

db.js
garethbowen: As below.

garethbowen:
@SCdF I'm not sure it's that simple because of the `module.exports = nano(baseUrl);` rubbish. That's presumably also the reason we export a completely different object when testing, which is another really bad code smell. I agree this needs to be fixed but I think it would involve changing the structure of the db object that's passed around so would be a large refactor.

I would argue for using this as further impetus to take the next step and replace nano with pouchdb.

db.js
SCdF: 😱

garethbowen:
This has been merged into medic-webapp manually due to submodule merging.

PR #179, opened 2017-10-31
Allow SMS to be sent to other people based on bool_expr

SCdF:
# Description

Two main changes:
 - Allow for you to set the recipient phone number to an arbitrary JS
   expression, given the doc as context
 - Allow for you to decide whether or not to send out an SMS based on
   an arbitrary JS expression, given the doc as context

medic/medic-webapp#3412

# Review checklist

### DIFF
--- a/lib/lineage.js
+++ b/lib/lineage.js
@@ -2,6 +2,10 @@ const _ = require('underscore'),
     db = require('../db'),
     utils = require('./utils');
 
+const findPatientId = doc =>
+  doc.type === 'data_record' &&
+  ((doc.fields && doc.fields.patient_id) || doc.patient_id);
+
 const fillContactsInDocs = (docs, contacts) => {
   if (!contacts || !contacts.length) {
     return docs;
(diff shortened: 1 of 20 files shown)


lib/lineage.js
```diff
@@ -47,6 +48,26 @@ const minifyContact = contact => {
   return result;
 };
 
+// TODO: optimise this so we don't have to recurse here
+//       Intergrate this tigher into fetchHydratedDoc (if we want to keep this)
+//       so we call docs_by_lineage twice, and then combine our contact fetch
+//       calls. This saves us one fetch call, and probably quite a few duplicate
+//       documents pulled. (It is probable that the contact hierarchy of the
+//       patient is the same or similar to the contact)
+const attachPatient = doc =>
```
SCdF: Still working on refactoring this out

lib/lineage.js
```diff
@@ -47,6 +48,26 @@ const minifyContact = contact => {
...
+  new Promise((resolve, reject) => {
+    utils.getPatientContactUuid(db, doc.fields.patient_id, function(err, uuid) {
+      if (err) {
+        reject(err);
+      } else {
+        fetchHydratedDoc(uuid).then(hydratedPatientDoc => {
```
garethbowen: What if `fetchHydratedDoc` errors? I _think_ you can just resolve a promise, so something like...

```js
} else {
  resolve(fetchHydratedDoc(uuid).then(hydratedPatientDoc => {
    doc.patient = hydratedPatientDoc;
    return doc;
  }));
}
```

Or maybe it's better to put the promise stuff outside of the Promise wrapper? eg:

```js
const attachPatient = doc =>
  new Promise((resolve, reject) => {
    utils.getPatientContactUuid(db, doc.fields.patient_id, function(err, uuid) {
      if (err) {
        reject(err);
      } else {
        resolve(uuid);
      }
    });
  })
  .then(fetchHydratedDoc)
  .then(hydratedPatientDoc => {
    doc.patient = hydratedPatientDoc;
    return doc;
  });
```

lib/lineage.js
```diff
@@ -91,7 +112,13 @@ const fetchHydratedDoc = id => {
           fillContactsInDocs(lineage, allContacts);
           const doc = lineage.shift();
           buildHydratedDoc(doc, lineage);
-          return doc;
+
+          // Also attach the patient if we should
+          if (doc.type === 'data_record' && doc.fields && doc.fields.patient_id) {
```
garethbowen: Might you also want to do this for registrations where the `patient_id` is generated not submitted and therefore not under the `fields` property?

lib/lineage.js
```diff
@@ -229,10 +257,11 @@ module.exports = {
    * @param docs The array of docs to hydrate
    * @returns Promise
    */
+   //TODO @gareth should I add patients to this?
```
garethbowen: Yeah probably best to be consistent about what "hydrated" means...

lib/utils.js
```diff
@@ -530,5 +531,9 @@ module.exports = {
   getPatientContact: (db, patientShortcodeId, callback) => {
     getPatient(db, patientShortcodeId, true, callback);
   },
+  isValidBooleanExpression: expr =>
+    typeof expr === 'string' && expr.trim() !== '',
+  evalExpression: (context, expr) =>
+    vm.runInNewContext(expr, context),
```
garethbowen: Switching the order of params is confusing to me - `expr`, `context` is the right order IMO.

lib/utils.js
```diff
@@ -530,5 +531,9 @@ module.exports = {
   getPatientContact: (db, patientShortcodeId, callback) => {
     getPatient(db, patientShortcodeId, true, callback);
   },
+  isValidBooleanExpression: expr =>
```
garethbowen: This name implies it does more than it does - really it's just checking that it's a string that's not empty. It could still be an invalid expression.

lib/messages.js
```diff
@@ -175,12 +175,21 @@ module.exports = {
...
+        } else if (doc.fields && doc.fields[recipient]) {
+            // Try to resolve a specified property/field name
             phone = doc.fields[recipient];
+        } else if (doc[recipient]) {
+            // Or directly on the doc
+            phone = doc[recipient];
+        } else if (recipient.indexOf('.') > -1) {
+            // Or multiple layers by executing it as a statement
+            try {
+                phone = utils.evalExpression({doc: doc}, 'doc.' + recipient);
```
garethbowen: I know we use `eval` everywhere but maybe we can come up with a safer way to walk a JSON object? For example we use [object-path](https://www.npmjs.com/package/object-path) in api.

lib/lineage.js
SCdF: I've raised #4003 so I can do this later in 2.15 and not be held back doing more important things

lib/messages.js
SCdF: Good idea!

lib/utils.js
SCdF: I was keeping the order that the things I was refactoring used. I'll switch it around

SCdF:
Closed because of merging repos. New equivalent is here: https://github.com/medic/medic-webapp/pull/4005

PR #180, opened 2017-11-21
Sentinel looks in the old location for its metadata file #4054

SCdF:
We used to have a migration that moves the location. However, that can
cause conflicts because Sentinel runs while API runs, including
migrations.

Instead, manage it inside sentinel itself.

At some point in the future we could delete this, once we're confident
everyone has upgraded past that document.

medic/medic-webapp#4054

# Review checklist

### DIFF
--- a/transitions/index.js
+++ b/transitions/index.js
@@ -419,11 +419,37 @@ const getMetaData = callback =>
       if (err.statusCode !== 404) {
         return callback(err);
       }
-      doc = {
-        _id: METADATA_DOCUMENT,
-        processed_seq: 0
-      };
+
+      // Doc doesn't exist.
+      // Maybe we have the doc in the old location?
+      return db.medic.get('sentinel-meta-data', (err, doc) => {
+        if (err) {
+          if (err.statusCode !== 404) {
+            return callback(err);
+          }
+
+          // No doc at all, create and return default
+          doc = {
+            _id: METADATA_DOCUMENT,
+            processed_seq: 0
+          };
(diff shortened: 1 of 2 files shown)


transitions/index.js
```diff
@@ -419,11 +419,43 @@ const getMetaData = callback =>
...
+
+          doc._id = METADATA_DOCUMENT;
+          delete doc._rev;
+          callback(null, doc);
+        });
+      });
     }
+
+    // Doc exists in correct location
     callback(null, doc);
```
alxndrsn: Could this be moved to the top of the function?  E.g.
```
if(!err) {
  // Doc exists in correct location
  return callback(null, doc);
}
```
I think this would make it much more obvious what's going on.

You could probably remove two levels of indents this way:
```
if(!err) {
  // Doc exists in correct location
  return callback(null, doc);
}
if(err.status !== 404) {
  return callback(err);
}
```
…handle 404…

transitions/index.js
```diff
@@ -419,11 +419,43 @@ const getMetaData = callback =>
...
+        };
+        logger.info('Deleting old metadata document', doc);
+        db.medic.insert(stub, err => {
+          if (err) {
+            return callback(err);
+          }
+
+          doc._id = METADATA_DOCUMENT;
+          delete doc._rev;
+          callback(null, doc);
```
alxndrsn: If this data is Very Important, it's probably safer to create the new document before deleting the old one.

PR #181, opened 2018-01-23
Guard against an empty patient

SCdF:
Just as we error if there is no registration, we should also error
if there is no patient. There should no longer be cases where there
are not patients attached.

medic/medic-webapp#4121

# Review checklist


SCdF:
Merged manually.

PR #183, opened 2018-05-05
Handle empty object parents in lineage

alxndrsn:
Issue: medic/medic-projects#3936

This may need porting to:
- [ ] `master`
- [ ] `2.15.x`

### DIFF
--- a/lib/lineage.js
+++ b/lib/lineage.js
@@ -35,7 +35,8 @@ const buildHydratedDoc = (doc, lineage) => {
     // if the parent doc isn't found the element in lineage is null,
     // so retain the stub.
     const parent = lineage.shift() || currentStub;
-    if (!parent) {
+    if (_.isEmpty(parent)) {
+      delete currentParent.parent;
       break;
     }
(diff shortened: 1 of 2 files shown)


lib/lineage.js
```diff
@@ -35,7 +35,8 @@ const buildHydratedDoc = (doc, lineage) => {
     // if the parent doc isn't found the element in lineage is null,
     // so retain the stub.
     const parent = lineage.shift() || currentStub;
-    if (!parent) {
+    if (_.isEmpty(parent)) {
```
garethbowen, line 38: LGTM.

lib/lineage.js
```diff
@@ -35,7 +35,8 @@ const buildHydratedDoc = (doc, lineage) => {
...
+      delete currentParent.parent;
```
garethbowen, line 39: I feel like this line fits better in the `minify` function. This is particularly true in the other branches where the shared library minify is called when creating the records too.

garethbowen (approved):

